PR #86, opened 2014-11-07
Make the return value of "Parse metadata" consistent

fmarier:
The output of this algorithm should be the same in all three
cases:

- the empty string
- a list of one or more metadata with unrecognized hash functions
- a list of one or more metadata with invalid ni URI

This change makes the last two cases match the output of the first.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -390,7 +390,8 @@ the user agent.
         <var>token</var>.
     3.  If <var>algorithm</var> is a hash function recognized by the user
         agent, add <var>token</var> to <var>result</var>.
-4.  Return <var>result</var>.
+4.  Return `no metadata` if <var>result</var> is empty, otherwise return
+    <var>result</var>.
 
 [split-on-spaces]: http://www.w3.org/TR/html5/infrastructure.html#split-a-string-on-spaces
 </section><!-- Algorithms::parse -->


devd:
I am not sure if the right thing is to return empty value from algorithm in all three cases or to treat all three as empty in the spec. The different values could allow for flexibility in the future, maybe? And I imagine UAs might want to implement that anyhow to differentiate warnings in the console. WDYT @mozfreddyb @metromoxie

metromoxie:
I don't feel too strongly either way. I think the point about warnings is good, but UAs could, of course, juts do that mid-algorithm anyway.

fmarier:
@devd Speaking for the Firefox implementation, the algorithm in the spec is not really what I've implemented anyways. I do check for various ways in which things fail to be valid.

However, I think these are implementation details and that the spec algorithm can omit these details and focus on being simple to understand and reason about.

mozfreddyb:
I hope that users and authors in our spec audience will outnumber us user agent people. So maybe we should optimize for clarity, not for implementation?

mikewest:
I'd agree that you should optimize for clarity, but I would tentatively suggest that the _prose_ in your specification is for authors and users, and it should explain enough for them to understand both the intent and impact of the feature. The algorithms are probably best written with implementers as the target audience.

devd:
lets merge this for now and then take another edit pass at it later. I imagine we will need to edit the use of ni:// URIs too.

PR #98, opened 2014-11-27
Mention the one thing that integrity protects against on insecure origins

fmarier:
It's not strictly true that there are no security benefits at all
for including integrity metadata on an insecure page.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -721,9 +721,10 @@ in general.
 origin against a compromise of the server where an external resources is
 hosted. Network attackers can alter the digest in-flight (or remove it
 entirely (or do absolutely anything else to the document)), just as they
-could alter the resource the hash is meant to validate. Authors who desire
-any sort of security whatsoever SHOULD deliver resources containing
-digests over secure channels.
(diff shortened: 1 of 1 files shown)


fmarier:
@metromoxie does that wording sound good to you?

devd:
I think instead of saying "any security whatsoever" we should say just say "Authors SHOULD deliver integrity metadata over HTTPS" and maybe reference https://w3ctag.github.io/web-https/ (since that is a broader finding and rec that we agree with)

fmarier:
I have revised the wording following @devd's suggestion.

devd:
minor nit, I would say "see also ..[securing the web]"

but other than that, lgtm. @metromoxie @mozfreddyb wdyt?

fmarier:
@devd Like this? https://github.com/w3c/webappsec/pull/98/files#diff-c3b7fb78ee74e440e18502426cd8b14fR725

devd:
yeah, this is great. lets wait for Joel/Freddy to chime in, if they want; otherwise I will merge later this week.

mozfreddyb:
It's an important notion, to distinguish between a server compromise and an active network attacker. Thanks for taking this! :+1:

metromoxie:
FWIW, lgtm :-)

PR #99, opened 2014-11-27
Move hyperlink auditing ("ping") to the connect-src directive

fmarier:
This change brings the CSP spec in line with the relevant section
of the Fetch standard:

  https://fetch.spec.whatwg.org/#requests

This is implementing the change that was proposed by @dveditz in this thread:

  http://lists.w3.org/Archives/Public/public-webappsec/2014Nov/0259.html


mikewest:
Merged per https://lists.w3.org/Archives/Public/public-webappsec/2015Jan/0271.html

PR #102, opened 2014-12-01
draft note for interaction of sandbox directive and workers

hillbrad:


mikewest:
I've taken your patch as a baseline, and poked at it a little in https://github.com/w3c/webappsec/commit/0d204ac3ede6fb48659efb11282909494fda756d

I hope it matches your intent!

Issue #68, opened 2014-10-28
SRI: noncanonical-src use cases


hillbrad:
"Issue-5" in the draft spec text states that noncanoncial-src only makes sense if we are allowing HTTP.  This is not true - it also allows for fallback in the case that e.g. a client is behind a proxy that intercepts and modifies HTTPS connections.  This is likely a nontrivial percentage of clients.  (e.g. enterprise and government systems)

fmarier:
The `noncanonical-src` attribute was removed in 3629390e542a5bc181740f64c5e4d6d6bf4f053c.

devd:
closing since we don't have noncanonical-src anymore

PR #118, opened 2014-12-22
Move the fallback / noncanonical bits to v2

fmarier:
As agreed during the 2014-12-15 teleconference
(http://www.w3.org/2014/12/15-webappsec-minutes.html#item07), the
fallback mechanism is moved to a future version.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -83,13 +83,6 @@ regardless of the URL from which they are loaded.
     libraries can be downloaded once, and only once, even if multiple
     instances with distinct URLs are requested.
 
-4.  (potentially) Allow resources to be downloaded from non-canonical
-    sources (for instance, over an insecure channel) for performance,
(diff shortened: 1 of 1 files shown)


mozfreddyb:
I suggest that Mike, Joel and Freddy of the future have to dig wildly through commit history if they want to resurrect these bits ;)

See also my previous commits of rage deletion in #74.

mozfreddyb:
(We will also be able to use github labels and milestones to see things that changed before we brought v1 into last call)

fmarier:
Once merged, #68 will no longer need to be fixed.

devd:
In the non-normative bits of the document, can we, instead of deleting outright, point out that this is a problem and developers should use other tricks to detect whether load was successful and fall back to a slower, non-cdn resource. 

(it is perfectly fine not to do it in this commit, but maybe leave a todo/issue in there when you delete those bits of text?)

fmarier:
@devd: rather then adding a TODO in the text, I've created #149 for it.

PR #120, opened 2014-12-23
Allow the empty set as a return value for parsemetada (fixes #119)

### LINKED ISSUE #119: Resolve ambiguity in the "parse metadata" algorithm between "()" and "no metadata"
The "parse metadata" algorithm in the spec currently return one of three things:
1. "no metadata"
2. an empty set
3. a non-empty set of integrity metadata

In #86, I argued that cases 1 and 2 should be the same and removed that distinction. There are two consequences to that change (or clarification):

- If a user agent sees a hash algorithm it doesn't support (e.g. `md5` or `sha-3000`), then it will ignore the integrity check and load the script, as per 3.3 of the parse metadata algorithm.
- If a user agent sees an integrity metadata with another scheme than `ni` (e.g. `bogus:///sha-256...`), then  it will ignore the integrity check and load the script, as per 3.1 of the parse metadata algorithm.

Is what I'm describing the intended behavior or do we intend to block both of these cases?

@mastahyeti uses the latter interpretation in his [SRI test page](https://btoews.github.io/sri_test/).

fmarier:
The consequence of this is that an integrity attribute with invalid
URIs or an unsupported hash will prevent the sub-resource from
being loaded.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -391,9 +391,9 @@ checking because it won't have loaded successfully.
 <section>
 #### Parse <var>metadata</var>.
 
-This algorithm accepts a string, and returns either `no metadata`, or a set of
-valid "named information" (`ni`) URLs whose hash functions are understood by
-the user agent.
+This algorithm accepts a string, and returns either `no metadata`, or a
+(potentially empty) set of valid "named information" (`ni`) URLs whose hash
+functions are understood by the user agent.
 
 1.  If <var>metadata</var> is the empty string, return `no metadata`.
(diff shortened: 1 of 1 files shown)


fmarier:
It seems like the consensus on the list was to fail closed, which means that we should consider this pull request and not #86.

devd:
hmm .. I am not convinced that was the consensus and I feel strongly enough about this that I have pinged the list again. lets see what comes up.

fmarier:
Based on recent discussions, I don't think we want to follow this approach since it would block the last 2 test cases from https://github.com/w3c/webappsec/issues/119#issuecomment-75689032 and we don't want to block test case 2.

PR #122, opened 2015-01-01
SRI: Use the {:.note} directive to mark note paragraphs up

fmarier:


fmarier:
I've rebuilt `index.html` using kramdown 1.5.0.

mozfreddyb:
Just to make sure: Did you just run kramdown or did you use the Makefile that comes with the spec, @fmarier?

PR #124, opened 2015-01-01
SRI: RFC6920 clarifications

fmarier:
We refer to RFC6920 but we should be more precise about the parts we're not using.


fmarier:
#156 may obsolete this pull request.

fmarier:
This is no longer needed since we removed NI URLs in #156.

PR #125, opened 2015-01-05
Add a step to normalize the hash function name (fix #104)

### LINKED ISSUE #104: normalize hash function algorithm name
As noted in #78, we should normalize the algorithm name:

> All ni URL syntax comes from RFC 6920 and on page 5, it keeps on repeating it as sha-256. Unfortunately, CSP2 regards writings like sha256 (no dash) as valid hashes.

> I'd prefer if everything was in line with the RFC, but the ship may have sailed for CSP2.
> So I propose we add normalization steps in SRI 3.3.3 Does resource match metadata?. Maybe we should do the same in CSP?

> Previously:
> 
>     5.Let algorithm be the alg component of metadata.
> 
> Suggested patch:
> 
>     5.Parse and normalize the algorithm from the alg component of metadata as follows
>     5.1.Stripping any dashes and transform all ASCII characters to lowercase ASCII.
>

fmarier:
This is the patch suggested by Freddy on https://github.com/w3c/webappsec/issues/104#issue-51271464.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -444,7 +444,9 @@ the user agent.
         skip the remaining steps, and proceed to the next token.
     2.  Let <var>algorithm</var> be the <var>alg</var> component of
         <var>token</var>.
-    3.  If <var>algorithm</var> is a hash function recognized by the user
+    3.  Strip any dashes from <var>algorithm</var> and transform all ASCII
+        characters to lowercase ASCII.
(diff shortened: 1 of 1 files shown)


devd:
this is great! I assume you are taking care of this fore Firefox. Flagging for @metromoxie @mikewest for Chrome.

mozfreddyb:
I guess @mikewest is right here. We should make things easier, but not too weird ;)

devd:
thanks! merged. sorry folks .. should have caught that!

PR #140, opened 2015-01-08
SRI: Prescriptivize usage of `ni` URLs for metadata.

mikewest:
This addresses https://www.w3.org/Bugs/Public/show_bug.cgi?id=27748


mikewest:
@mozfreddyb WDYT?

mozfreddyb:
:+1:

PR #143, opened 2015-01-09
Remove the anchor element from v1 of the spec

fmarier:


devd:
Was there a decision on this on the list? Should we just bring this up on teh call? Are you definitely not covering this in the Firefox implementation? What's happening in the Chrome implementation @metromoxie  ?

fmarier:
The Firefox and Chrome implementations are currently limited to the `script` element.

I'm planning to add stylesheet support in Firefox, but not the `a` element.

I'm told removing this was discussed at TPAC, but I don't think it was brought up on the list, so perhaps it should be discussed during the call.

devd:
it was discussed as an option, but not decided. lets see what @metromoxie can say about the status.

@hillbrad can we discuss it in the call tomorrow?

metromoxie:
We do not implement it in Chromium, nor do we have a plan to. I could me misremembering, but as @fmarier says, I thought we decided to punt it to V2 at TPAC. We should probably discuss one more time on the call though.

specs/subresourceintegrity/spec.markdown
```diff
@@ -310,18 +297,17 @@ resource in order to provide agility in the face of future discoveries.
 For example, the "Hello, world." resource described above may be described
 either of the following `ni` URLs:
 
-    ni:///sha-256;-MO_YqmqPm_BYZwlDkir51GTc9Pt9BvmLrXcRRma8u8?ct=text/plain
-    ni:///sha-512;rQw3wx1psxXzqB8TyM3nAQlK2RcluhsNwxmcqXE2YbgoDW735o8TPmIR4uWpoxUERddvFwjgRSGw7gNPCwuvJg?ct=text/plain
+    ni:///sha-256;-MO_YqmqPm_BYZwlDkir51GTc9Pt9BvmLrXcRRma8u8?ct=application/javascript
+    ni:///sha-512;rQw3wx1psxXzqB8TyM3nAQlK2RcluhsNwxmcqXE2YbgoDW735o8TPmIR4uWpoxUERddvFwjgRSGw7gNPCwuvJg?ct=application/javascript
 {:.example.highlight}
```
fmarier: Good point. I've created #144 for this since it's a bigger change than just removing support for the anchor element. Feel free to assign it to me.

metromoxie:
From the last conference call, it sounded like there was agreement that we should *not* include <a> in v1. Is my understanding correct?

devd:
Unfortunately, I had to unexpectedly miss that call. If that's what the
everyone agreed on then sgtm

fmarier:
The discussion was short but nobody seemed to have any objections to moving it to v2: http://www.w3.org/2015/01/12-webappsec-minutes.html#item09

devd:
thanks

PR #148, opened 2015-01-18
Remove all references to fonts from the spec (fix #17)

### LINKED ISSUE #17: [SRI] Talk about Fonts
(HT Tanvi)

The spec needs to talk about fonts. It mentions it in the beginning but never follows up with details.

fmarier:
Since v1 will not cover fonts, we should avoid mentioning them in
the spec.


devd:
thanks!

PR #152, opened 2015-01-19
Remove section about XHRs for v1

fmarier:
I don't recall this being discussed on the list or on the call, so maybe we should bring it up during the next teleconf.

Mozilla is not planning on implementing this in Firefox. What about Chromium, @metromoxie ?


metromoxie:
We have not implemented XHR integrity and do not plan to. Let's remove.

Issue #147, opened 2015-01-16
[CSP] Accept base64url values for script hash


metromoxie:
The Subresource Integrity draft uses base64url and, in practice, Chrome's CSP implementation accepts both base64 and base64url values. I propose that we standardize that both are acceptable.

metromoxie:
FYI, I started a discussion on public-webappsec: https://lists.w3.org/Archives/Public/public-webappsec/2015Jan/0156.html

PR #154, opened 2015-01-20
Separate base64-value and base64url-value

metromoxie:
As a follow up to w3c/webappsec#147 and 7b3b425ae340734a088323ff2ffb7f575d7e3163, I thought it might be nice to syntactically separate out base64 values and base64url values. @mikewest, what do you think?


mikewest:
LGTM

PR #156, opened 2015-01-27
Adopt the same format as CSP Level 2 for the hash format

fmarier:
This is a first cut for what the changes proposed in this thread might look like:

  https://lists.w3.org/Archives/Public/public-webappsec/2015Jan/0177.html

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -35,7 +35,7 @@ the representation of the resource the author expects to load. For instance,
 an author may wish to load jQuery from a shared server rather than hosting it
 on their own origin. Specifying that the _expected_ SHA-256 hash of
 `https://code.jquery.com/jquery-1.10.2.min.js`
-is `C6CB9UYIS9UJeqinPHWTHVqh_E1uhG5Twh-Y5qFQmYg` means
+is `C6CB9UYIS9UJeqinPHWTHVqh/E1uhG5Twh+Y5qFQmYg=` means
 that the user agent can verify that the data it loads from that URL matches
 that expected hash before executing the JavaScript it contains. This
 integrity verification significantly reduces the risk that an attacker can
@@ -45,7 +45,9 @@ This example can be communicated to a user agent by adding the hash to a
 `script` element, like so:
 
     <script src="https://code.jquery.com/jquery-1.10.2.min.js"
-            integrity="ni:///sha-256;C6CB9UYIS9UJeqinPHWTHVqh_E1uhG5Twh-Y5qFQmYg?ct=application/javascript">
+            integrity="type:application/javascript
+                       sha256-C6CB9UYIS9UJeqinPHWTHVqh/E1uhG5Twh+Y5qFQmYg=">
+
 {:.example.highlight}
 
 Scripts, of course, are not the only resource type which would benefit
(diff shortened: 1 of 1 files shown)


fmarier:
This is **not** meant to be merged now, it's just meant to restart the conversation around the format of the `integrity` attribute: https://lists.w3.org/Archives/Public/public-webappsec/2015Jan/0259.html

fmarier:
It looks like CSP2 just [added base64url as an option](https://github.com/w3c/webappsec/pull/154) so we should double-check this before merging this PR, should we decide that it's the right approach.

metromoxie:
We added base64url as an option to CSP because (a) we wanted to be consistent with the SRI spec, (b) we felt that there was no reason to force developers into one format or the other, and (c) we figured requiring one of the other would lead to confusing errors where your hash was correct but just encoded wrong.

I think we should add base64 as a viable encoding for digests in SRI. Should I open an issue for that?

fmarier:
@metromoxie I've just created #163 for this.

fmarier:
I've just updated this PR to reflect my latest proposal from https://lists.w3.org/Archives/Public/public-webappsec/2015Jan/0340.html

fmarier:
@metromoxie @devd @mozfreddyb are you guys happy with this proposal? is there anything you'd like me to change?

The feedback from the list seems to have stopped, so I think we can move forward with this change. Nobody seems to like NI URIs...

devd:
I like the move away from ni URIs but I just dislike the type argument in the attribute :(; but I can't think of anything better. I think we should just kill the type attribute and just infer defaults like the rest of the web.

what do others think?

hillbrad:
Probably there are not many practical attacks.  I was concerned with things
like a chimeric script that is interpreted to do one thing as
text/javascript and something quite different as text/vbscript, or things
like GIFARs.  But I'm having a difficult time right now imagining a full
exploit kill-chain  using this kind of technique which SRI would be likely
to interrupt.

devd:
Since removing the type will match CSP L2 hash format completely, maybe we can do that as part of this bug? ;) @fmarier

mikewest:
The type is fairly essential if we want to work with caching improvements later on (e.g. we need to know that the resource should be loaded as an image, and not as a script, etc. We note this briefly in http://www.w3.org/TR/SRI/#mime-type-confusion-1. I'd prefer keeping it in, in the hopes that folks start using it now as a defense-in-depth (e.g. it will encourage them to deliver resources with reasonable MIME types, which will also help browsers that don't support SRI).

devd:
hmm .. the current wording seems to suggest it is not mandatory to specify the type in the integrity metadata; does that match your understanding? If that's the case, I am fine with that.

Separately, I am confused why we need the type for caching once we have the Access-Control-Allow-origin * on a resource, it is anyhow world readable. What does the extra type check help protect us from? (maybe that should be a separate issue?)

fmarier:
I have opend #176 to follow-up on whether or not we want to keep content-type in V1.

Since we agree on matching the CSP2 hash format, I would like to suggest that we merge this PR and then follow-up with another that removes the content-type stuff if that's what we decide in #176.

Given that the presence of content-type is a little controversial, we should avoid bundling its removal with the hash format change (this PR) and instead discuss it some more on list and/or teleconf.

devd:
good point. thanks for writing this!

PR #158, opened 2015-01-28
Stylesheet links without a MIME type should be eligible for integrity checks

fmarier:
HTML allows stylesheets to be loaded without specifying a MIME type of text/css.

We should encourage developers to supply the MIME type but we should not require it before enforcing integrity checks on it.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -631,7 +631,7 @@ failure, refuse to fetch the resource, and [report a violation][].
 ###### The `link` element for stylesheets
 
 Whenever a user agent attempts to [obtain a resource][] pointed to by a
-`link` element that has a `rel` attribute with the value of `stylesheet` and a type of `text/css`:
+`link` element that has a `rel` attribute with the value of `stylesheet`:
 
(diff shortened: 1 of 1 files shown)


mozfreddyb:
I'm not sure about this.

We need a MIME type in the integrity check to avoid ambiguity with files that are valid to multiple parsers (think: GIFAR).

If we end up changing the integrity syntax away from `ni://` URLs to something else, we might want the type from the attribute.

fmarier:
It's certainly better to include the MIME type in the integrity attribute, but we don't currently require it for `<script>` elements so it feels odd to require it just for `<link>` elements.

metromoxie:
@mozfreddyb Is this any worse than the status quo though? I don't think this makes things any worse, and it feels funny to use SRI as a place to enforce MIME types.

fmarier:
If we're going to enforce MIME types (which I'm not saying we should), we should do so within the integrity attribute IMHO.

devd:
+1 to Joel. I don't think SRI should get into the mime-types mess.

mozfreddyb:
Oh yes. You're right. *merging*

PR #161, opened 2015-01-29
CSP3: message-src message-sink

deian:
First cut at the message-src and message-sink directives. Fixed references to compile with bikeshed.

I tried to reopen #101 by pushing to master with no avail.

(Sorry for being MIA for the last few weeks. And, thanks for closing the other issues!)

### DIFF
--- a/specs/content-security-policy/index.src.html
+++ b/specs/content-security-policy/index.src.html
@@ -178,6 +178,8 @@ type: interface
       text: HTMLStyleElement
     urlPrefix: scripting-1.html
       text: HTMLScriptElement
+    urlPrefix: browser.html
+      text: Window
   urlPrefix: http://www.w3.org/TR/dom/
     text: Node; url: interface-node
     text: Event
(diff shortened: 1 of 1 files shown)


mikewest:
Please feel free to start a thread on public-webappsec@ to see what folks think. I'm trying to focus on CSP2 for the next week or two just to get it out the door, but this looks like a totally reasonable thing to add to CSP3. Thank you!

devd:
[at a glance]

it seems that the wording needs a bit of a clarification on how it handles messageChannel based messaging and post message between workers and pages, workers and other workers etc. As far as I can tell, this is currently only meant to apply to documents, which is fine but clarifying that will be helpful.

deian:
Thaks! @devd this is actually something I wanted to discuss on the list/here. Channels do not have origins associated with them, so this is a bit tricky. See https://lists.w3.org/Archives/Public/public-webappsec/2015Jan/0339.html for a discussion with Brad on this. I am inclined to say that we should handle the messageChannel-based messaging, but am interested to hear what others think. Any thoughts on this?

devd:
I guess I am a bigger fan of MVP in the first stage and seeing if web
applications find it useful instead of solving all problems in the very
first spec. Restricting to just iframes (i.e., origin-based) messaging
seems like a great first step and we can move to channel messaging when
there is an ask.

deian:
That sounds like a reasonable first step forward to me. I'll update the text accrodingly.

Malvoz:
I must assume CSP won't need this directive when https://github.com/whatwg/html/issues/3740 has been resolved.

devd:
Hmm.. maybe I am missing something but doesn't look like it lets you limit
destinations or sources. It allows you to block postmessage outright which
is useful but not as powerful as csp

annevk:
This is superseded by https://github.com/w3c/webappsec-csp/issues/117. (CSP issues need to go against the CSP repository these days.)

PR #167, opened 2015-02-05
Fix two typos

foolip:


mikewest:
I _really_ need to figure out how vim's spell checker works. Thanks!

PR #171, opened 2015-02-09
CSP2: remove base64url and sha-*

fmarier:
As per https://lists.w3.org/Archives/Public/public-webappsec/2015Feb/0153.html, I have reverted the changes that were made to bring CSP in line with the SRI spec.


mikewest:
LGTM, thanks!

PR #172, opened 2015-02-10
minor typos, seriousing-stuff-up for FPWD

hillbrad:


specs/csp-pinning/index.src.html
```diff
@@ -97,10 +97,14 @@ <h3 id="use-cases">Use Cases</h3>
                                style-src https://application2.cdn.com;
     </pre>
 
+		ISSUE: This example contradicts advice elsewhere that the
```
mikewest: Are you using tabs? Or some other weird spacing? :)

mikewest:
Thanks!

specs/csp-pinning/index.src.html
```diff
@@ -97,10 +97,14 @@ <h3 id="use-cases">Use Cases</h3>
...
+    value of the header SHOULD be the same for every <a>resource
+    representation</a>.
```
mikewest: Looking again, I don't understand this issue. The pinned header _is_ the same in both examples, isn't it?

mikewest: Ah, one line was different. Fixed, thanks!

PR #173, opened 2015-02-10
Probably a bikeshed-ism, but let's not have the ??? in the biblio.

hillbrad:


mikewest:
This is a Bikeshedism. I'll file a bug.

PR #175, opened 2015-02-12
Add a new "referrer attribute" delivery mechanism

fmarier:
This delivery mechanism is meant to subsume the rel="noreferrer"
link type defined in the HTML5 spec:

  https://html.spec.whatwg.org/multipage/semantics.html#link-type-noreferrer

This would allow for the following use case:

- I don't want any referrers to be sent (i.e. global policy of
"no-referrer" delivered via <meta> or CSP)
- except for these two links that should reveal the origin of the
referring page (i.e. local policy of "origin" on the two <a> elements)

### DIFF
--- a/specs/referrer-policy/index.html
+++ b/specs/referrer-policy/index.html
@@ -71,7 +71,7 @@
    <h1 class="p-name no-ref" id="title">Referrer Policy</h1>
   
    <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft,
-    <time class="dt-updated" datetime="2015-02-10">10 February 2015</time></span></h2>
+    <time class="dt-updated" datetime="2015-02-15">15 February 2015</time></span></h2>
   
    <div data-fill-with="spec-metadata">
     <dl>
@@ -83,6 +83,8 @@
      <dd><a href="https://github.com/w3c/webappsec/commits/master/specs/referrer-policy/index.src.html">https://github.com/w3c/webappsec/commits/master/specs/referrer-policy/index.src.html</a>
      <dt>Feedback:
      <dd><span><a href="mailto:public-webappsec@w3.org?subject=%5BREFERRER%5D%20feedback">public-webappsec@w3.org</a> with subject line “<kbd>[REFERRER] <var>… message topic …</var></kbd>” (<a href="http://lists.w3.org/Archives/Public/public-webappsec/" rel="discussion">archives</a>)</span>
+     <dt>Issue Tracking:
+     <dd><a href="#issues-index">Inline In Spec</a>
      <dt class="editor">Editors:
      <dd class="editor">
       <div class="p-author h-card vcard"><a class="p-name fn u-email email" href="mailto:eisinger@google.com">Jochen Eisinger</a> (<span class="p-org org">Google Inc.</span>)</div>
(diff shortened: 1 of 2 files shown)


fmarier:
@mikewest I think I've addressed all of the feedback from the list.

annevk:
I think we should not do this and pursue https://www.w3.org/Bugs/Public/show_bug.cgi?id=26533 instead. We cannot keep request variables everywhere forever.

mikewest:
Well, either way we'd need to implement it as a parameter to Fetch, right? The question is whether we'd allow all of those parameters to live together in a single attribute (which makes sense to me).

I don't see any problem landing this to signal the direction we want to work in, iff we add another issue to the doc pointing at that bug, and work with Hixie, et al. to get whatever we end up with added to HTML.

briansmith:
Do you mean that instead of having a referrer attribute, it should be part of some JSON blob inside a src-options attribute?

If so, I don't think that is a good idea at all. 1. src-options being a JSON string means that it becomes yet another attack vector for XSS (you have to JSON escape things, and then attribute escape them). 2. For authors, the referrer attribute is much better ergonomically, and web developer ergonomics is more important than specification or implementation ergonomics (this is in the HTML design principles). For those reasons, I expect the src-options stuff is going to get a lot of pushback from anywhere it will tried to be applied to, not just referrer.

We should block progress on the referrer spec on src-options. Per-element referrer control is blocking other important improvements to the referrer spec. We should go ahead and add the referrer attribute to the editor's draft now, and if/when src-options becomes a preferable alternative, we can switch the spec to that assuming that happens in a reasonable timeframe.

fmarier:
I've addressed @briansmith's [concerns about the precedence](https://github.com/fmarier/webappsec/commit/3c7ad6d8bbaf47567e9bd3b580f0d469ab4bf69c#commitcomment-9742303) by turning that paragraph into an issue.

I have also pruned the list of allowed values based on [his comment to that effect](https://github.com/fmarier/webappsec/commit/406a65a7f4ccc58472fb0a493d3cb0c7944d3532#commitcomment-9742435).

specs/referrer-policy/index.html
```diff
@@ -961,14 +972,72 @@ <h3 class="heading settled" data-level="4.2" id="referrer-policy-delivery-meta">
...
+  <code><a data-link-type="element" href="https://html.spec.whatwg.org/#the-img-element">img</a></code> or <code><a data-link-type="element" href="https://html.spec.whatwg.org/#the-iframe-element">iframe</a></code>
+  element, for example:</p>
+
+  
+    <pre class="example">&lt;a href="http://example.com" referrer="origin">
+</pre>
+
+
+    <p>The following values for the <code>referrer</code> attribute are valid,
+  and map to the listed <a data-link-type="dfn" href="#referrer-policy">referrer policy</a> values:</p>
```
sc0ttbeardsley: Is there any reason to limit to just these three? It might make sense to support all of the following here: no-referrer, no-referrer-when-downgrade, origin, origin-when-cross-origin, unsafe-url

I only need the three listed at the moment.

sc0ttbeardsley:
I agree with @briansmith regarding https://www.w3.org/Bugs/Public/show_bug.cgi?id=26533 It seems like a lot of work which will be prone to error to specify this in a JSON blob.

I would like to see this PR land as it meets my needs for being able to override the per-document global policy for just a few links. LGTM

mikewest:
We'll run with it for the moment, and see if the other bug resolves in a way that folks are happy with. If not, `referrer` it is. :)

Thanks, @fmarier!

WooseokJeong:
Hi, would you confirm that this will support hierarchical inheritance and override in DOM structure?
`<div referrer='origin'>
<a href='http://sample.com/for/unsafe-url' referrer='unsafe-url'> 
<a href='http://sample.com/for/origin'> 
</div>`

fmarier:
@WooseokJeong It works kinda like that except not with a `<div>`. You can set a default policy for the whole document (via the `<meta>` tag) and then you can override it on a per-link basis.

WooseokJeong:
got it, thanks
I thought it might be good because we can control policies across element section level which also can reduce manipulation of every single links to override.

PR #178, opened 2015-02-18
Re-generate the index.html file

fmarier:
I've re-run `make` to pull in all of the recent changes to the spec.

@devd or @mozfreddyb what's the next step to push that to the github-hosted files?

### DIFF
--- a/specs/subresourceintegrity/index.html
+++ b/specs/subresourceintegrity/index.html
@@ -23,8 +23,8 @@
         // editors, add as many as you like
         // only "name" is required
         editors:  [
-            { name: "Frederik Braun", url: "mailto:fbraun@mozilla.com", company: "Mozilla Corporation", companyURL: "https://www.mozilla.com/" },
-            { name: "Francois Marier", url: "mailto:francois@mozilla.com", company: "Mozilla Corporation", companyURL: "https://www.mozilla.com/" },
+            { name: "Frederik Braun", url: "mailto:fbraun@mozilla.com", company: "Mozilla", companyURL: "https://www.mozilla.org/" },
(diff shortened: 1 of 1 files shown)


devd:
you need to push to the gh-pages branch typically, but I am surprised Mike doesn't have a script to do this already.

fmarier:
Thanks. I've cherry-picked the commit onto the `gh-pages` branch and pushed.

It would be nice if we only kept the generated files on the `gh-pages` branch. It would make it much easier to cherry-pick and revert commits on the the specs without having to deal with conflicts in `index.html` files.

devd:
since I only touch 1 spec and @mikewest touches all of them, I defer to him ;)

mikewest:
I've just been pushing master to gh-pages (`git push origin master master:gh-pages`). I don't think cherry-picking is a worthwhile use of our collective time. :)

mikewest:
In fact, looking at the diff now, this approach just makes things difficult. :( I'd prefer that we not intentionally allow `origin/master` and `origin/gh-pages` to diverge. I don't think there's any good reason to keep them separate, and doing so makes publication significantly more difficult.

You can simply run `make publish` from the `specs/` directory to publish to both `master` and `gh-pages`. I'd appreciate it if we could run with that.

mozfreddyb:
We should just push master *as is* to gh-pages, to keep workflows simple.

I've done merging and cherry-picking between master & gh-pages before, and it's indeed not a great way to spend one's time ;)

fmarier:
Ah sorry for the cherry-picking. It wasn't immediately obvious to me how you actually updated that branch. I hadn't realized it was just a copy of master.

PR #181, opened 2015-02-19
Change one of the examples to a stylesheet load

fmarier:
Since v1 includes support for both `<script>` and `<link rel="stylesheet">`
elements, we should include at least one example of both.


fmarier:
I have reverted the controversial HTTPS to HTTP change. This should be ready to be merged now.

PR #186, opened 2015-02-27
it's two header fields, right?

dret:


mikewest:
For the moment, yes. I think we're going to end up dropping the `report-only` variant, but let's merge this for now. Thanks for looking!

dret:
good to know, thanks for commenting. i'll keep an eye out for this in the next versions. i am trying to track the evolving landscape of HTTP headers.

mikewest:
Great. Are you summarizing the results anywhere? I'm curious. :)

dret:
HTTP header field summary is here: https://github.com/dret/sedola/blob/master/MD/headers.md
the new draft is described here: https://github.com/dret/sedola/blob/master/W3C/csp-pinning.xml
basically, the idea is to document the "web surface" of standards and services, so that it becomes easier to describe what a consumer should understand when it wants to use a service. it's a work in progress...

PR #190, opened 2015-03-02
Update examples with crossorigin=anonymous

devd:
For security, resources need to be CORS fetches or have public caching headers. I think it is better to have our examples use CORS fetches explicitly.

### DIFF
--- a/specs/subresourceintegrity/index.html
+++ b/specs/subresourceintegrity/index.html
@@ -148,7 +148,7 @@ has the ability to inject arbitrary content.</p>
 TLS, <a href="http://tools.ietf.org/html/rfc6797">HSTS</a>, and <a href="http://tools.ietf.org/html/draft-ietf-websec-key-pinning">pinned public keys</a>, a user agent can be fairly certain
 that it is indeed speaking with the server it believes it’s talking to. These
 mechanisms, however, authenticate <em>only</em> the server, <em>not</em> the content. An
-attacker (or admin!) with access to the server can manipulate content with
+attacker (or administrator) with access to the server can manipulate content with
 impunity. Ideally, authors would not only be able to pin the keys of a
 server, but also pin the <em>content</em>, ensuring that an exact representation of
 a resource, and <em>only</em> that representation, loads and executes.</p>
@@ -169,13 +169,13 @@ substitute malicious content.</p>
 <code>script</code> element, like so:</p>
 
   <pre><code>&lt;script src="https://code.jquery.com/jquery-1.10.2.min.js"
-        integrity="sha256-C6CB9UYIS9UJeqinPHWTHVqh/E1uhG5Twh+Y5qFQmYg="&gt;
+        integrity="sha256-C6CB9UYIS9UJeqinPHWTHVqh/E1uhG5Twh+Y5qFQmYg="
+        crossorigin="anonymous"&gt;
 </code></pre>
 
   <p>Scripts, of course, are not the only resource type which would benefit
-from integrity validation. The scheme specified here applies to all HTML
-elements which trigger fetches, as well as to fetches triggered from CSS
-and JavaScript.</p>
+from integrity validation. The scheme specified here also applies to <code>link</code>
+and future versions of the specification are likely to expand this coverage.</p>
 
   <section>
     <h3 id="goals">Goals</h3>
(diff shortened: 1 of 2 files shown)


mozfreddyb:
Looks good.

I think we should add a tiny ref to [https://html.spec.whatwg.org/multipage/semantics.html#attr-link-crossorigin](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-crossorigin)?

devd:
well, two concerns: where to add it and in general, we don't add a reference for html tags/attributes, right?

fmarier:
Given that we only support `<script>` and `<link>` elements, neither of which are prevented from loading cross-origin resources, why do we need to fetch using CORS?

fmarier:
Or does the credential-dropping option (`anonymous`) also apply to non-CORS fetches?

devd:
it is not actually cors fetch until we put that attribtue. So, with that attribute, teh Access-Control-Allow-Origin flag is checked and then when the right one is returned, everything works. So for anonymous, ACAO * will need to be returned. Otherwise, the CORS fetch will fail (but you can load the script without CORS; just not read it)

mozfreddyb:
I'd hate to will break the CDN use case for JS files that do not come with an ACAO *.

This does not execute:
> `data:text/html,<script src="http://people.mozilla.org/~fbraun/files/alert.js" crossorigin=anonymous></script>`

But this here does:
>  `data:text/html,<script src="http://people.mozilla.org/~fbraun/files/alert.js"></script>`

devd:
well, that will be caught in basic testing. The CDN only needs to send the ACAO *.

So, if you do a CORS fetch, you need to send ACAO *. This is the best solution.

Else, you don't do crossorigin=whatever fetch, and then it has to be same origin (or in the CDN case) publicly cacheable (but run time error catching might be trickier). Given that, I would rather that the spec encourages the ACAO * behavior since that is a pretty safe header to set.

mozfreddyb:
You're completely right for big CDNs. They almost always already send ACAO *.
But for those CDNs which don't, it's hard to convince them to enable it - if you're not a paying customer.

It's even harder to include something from a library on `gh-pages`.

In essence, this means that SRI won't work for "arbitrary JS file I found on the net and want to make sure doesn't change". :-/

annevk:
@mozfreddyb yeah, because that means violating SOP.

@devd what do you mean by public caching headers? That doesn't sound right.

annevk:
(Violating SOP in the sense that SRI can be used to read content if you're determined or if the content is small.)

mozfreddyb:
Well according to the *[Is resource eligible for integrity validation](http://www.w3.org/TR/SRI/#is-resource-eligible-for-integrity-validation)* algorithm, we don't need CORS to perform an integrity check. Explicitly, we need just either of these three:

> If the mode of request is CORS, return true.
> If the origin of request is resource’s origin, return true.
> If resource is [cachable by a shared cache](https://svn.tools.ietf.org/svn/wg/httpbis/draft-ietf-httpbis/latest/p6-cache.html#response.cacheability), as defined in [HTTP11], return true.

I don't think the cross-domain data can be considered sensitive, if it is publicly cachable. A "shared cache" is a proxy server, right?

NB: 
This algorithm describes the consensus we have achieved on the mailing list in the "Subresource Integrity and fingerprinting" thread (most notably [Michal Zalewski's email](https://lists.w3.org/Archives/Public/public-webappsec/2014Jan/0025.html) and @devd's [response](https://lists.w3.org/Archives/Public/public-webappsec/2014Jan/0026.html)).

annevk:
So far we have not allowed reading intranet content when it set `Cache-Control: public`. It's not entirely clear to me why we'd start now. Sure there was that email, but what changed about SOP?

The response is also somewhat puzzling as one of the networking rules is that we treat every violation the same as a network error to not give attackers more information than strictly required.

mozfreddyb:
Intranets ruin everything :suspect:

Seems like we have to revise that algorithm...

mozfreddyb:
:+1: for adding `crossorigin=anonymous` to the requests then

fmarier:
Does the fact that we don't trigger `onerror` unless the origin is the same change this reasoning at all?

In the example of an external site loading a publicly cachable intranet resource, we perform the SRI check, but we pretend everything is fine on a bad hash (while cancelling the load): https://w3c.github.io/webappsec/specs/subresourceintegrity/#the-script-element

devd:
@annevk The idea of using public-caching was something that was decided on in the mailing lists. fwiw, I am not a fan of that and I prefer the explicit opt-in of CORS. Since Freddy has filed #207 maybe we can merge this PR? The question of whether or not to use public caching as an opt-in for SRI is something I will bring up on the list.

(please let me bring it up since right now it won't make sense until I present all the other changes we made.)

annevk:
@devd I get that it was decided, but it is not really acceptable. This PR makes total sense though :+1:

devd:
@fmarier what do you think about merging this? We can discuss the issues with publicly cacheable resources in #207

mastahyeti:
> You're completely right for big CDNs. They almost always already send ACAO *.
But for those CDNs which don't, it's hard to convince them to enable it - if you're not a paying customer.

> It's even harder to include something from a library on gh-pages.

GitHub Pages now sends ACAO * for all responses :smile:.

metromoxie:
This lgtm, although you'll need a rebase because of the changes with types/options.

fmarier:
I've just rebased this for @devd. I think it's ready to go, but I'll let someone else double-check before merging.

metromoxie:
Two comments:
  * Can you add "crossorigin" to the one <link> example?
  * Would you mind compiling and adding the index.html?

Otherwise lgtm.

fmarier:
@metromoxie done

metromoxie:
Awesome, thanks. Merging.

devd:
awesome! thanks everyone

PR #193, opened 2015-03-02
Update example to use javascript and alert

### LINKED ISSUE #144: SRI: "Hello, world." should be replaced with a more appropriate example
Since the SRI spec now focuses on the `script` element, the current example hash is not particularly relevant, as [pointed out](https://github.com/w3c/webappsec/pull/143#discussion_r22707058) by @mozfreddyb.

I suggest hashing something like "alert('Hello, World.');" instead.

devd:
Fixes #144

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -211,17 +211,17 @@ risk of certain attack vectors.
 This metadata MUST be encoded in the same format as the `hash-source`
 in [section 4.2 of the Content Security Policy Level 2 specification][csp2-section42].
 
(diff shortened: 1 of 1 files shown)


fmarier:
If I run `echo -n "alert('Hello, world.');" | openssl dgst -sha256 -binary | openssl enc -base64 -A`, I get:

> qznLcsROx4GACP2dm0UCKCzCG+HiZ1guq6ZZDob/Tng=

Did you accidentally drop the first character?

devd:
this is why we have reviews!

devd:
(fixed, thanks!)

PR #195, opened 2015-03-02
[CredentialManagement] Removed a step from Request a Credential with options algorithm

jmajnert:
I think this step was left in the algorithm by mistake. It doesn't make sense anyway.

There's a significant difference in wording between the "security" steps in "Request a Credential with options" and "Save credential" algorithms. The commit here is just a cleanup, this issue needs more attention.

### DIFF
--- a/specs/credentialmanagement/index.src.html
+++ b/specs/credentialmanagement/index.src.html
@@ -755,15 +755,10 @@ type: constructor
       <strong>and</strong> terminate this algorithm.
     </li>
     <li>
-      If <var>settings</var>' <a>responsible document</a> is not the <a>active
-      document</a> in the <a>top-level browsing context</a>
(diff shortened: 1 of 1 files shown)


jmajnert:
This is also of importance here: https://github.com/w3c/webappsec/issues/197

mikewest:
Ugh, yes. That looks like a bad copy/paste at some point in the past. The requirement isn't an error, however; I don't intend for the API to be usable outside a top-level browsing context (e.g. `example.com` shouldn't be able to request credentials when it's not what the user sees in the address bar). I'm addressing this in a separate patch.

Regarding the difference in text between the algorithms, I've cleaned it up a bit, but I welcome any suggestions you might have regarding the checks we ought to make, and when.

PR #196, opened 2015-03-02
[CredentialManagement] Use of singular they instead of she

jmajnert:
I think that using the feminine pronouns in place of "user" noun is not quite correct English. Non English native-speakers might even be mislead into thinking that the noun user is of feminine gender.

The commit below changes most (all) such occurrences to use the "singular they" pronoun forms.


mikewest:
Hrm. This was an intentional choice on my part, and it's one that I don't think I want to change. I very much appreciate the PR, but I'd prefer to leave things they way they are on this point. :)

jmajnert:
No problem.

PR #201, opened 2015-03-03
remove require-for-all for v1

devd:
the require-for-all directive is a little brittle and not sure how it will play out with new resources and I am also not sure blocking when require-for-all is on but policy is report (issue 7 in the note) actually makes any sense.


fmarier:
makes sense to me, v+1 feature.

PR #202, opened 2015-03-03
CREDENTIAL: Fix for "top-level browsing context" dfn href

jmajnert:
The "-" was converted to "_" in href anchor.

Please first check with bikeshed if this solves the problem


mikewest:
Yeah, this is a problem I always forget about. Bikeshed distinguishes between `top level` and `top-level` when linkifying by converting the latter to `top_level`. The fix in your patch is the correct one. :)

PR #203, opened 2015-03-03
CREDENTIAL: Unified security error conditions

jmajnert:
Small cleanup. Hopefully will make the text clearer

### DIFF
--- a/specs/credentialmanagement/index.src.html
+++ b/specs/credentialmanagement/index.src.html
@@ -740,27 +740,24 @@ type: constructor
       Return <var>promise</var>, and execute the remaining steps asynchronously.
     </li>
     <li>
-      Let <var>environment</var> be the <a>JavaScript global environment</a> in
-      which which this algorithm is executing.
-    </li>
-    <li>
       Let <var>settings</var> be the <a>incumbent settings object</a>.
     </li>
     <li>
-      If <var>settings</var> does not have a <a>responsible document</a>, then
-      reject <var>promise</var> with a <code>SecurityError</code>
-      <strong>and</strong> terminate this algorithm.
-    </li>
-    <li>
-      If <var>settings</var>' <a>responsible document</a> is not the <a>active
-      document</a> in the <a>top-level browsing context</a>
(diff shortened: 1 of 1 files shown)


specs/credentialmanagement/index.src.html
```diff
@@ -747,20 +747,24 @@ <h3 id="request-credential">
...
+        </li>
+        <li>
+          <var>settings</var>' <a>responsible document</a> is not the
+          <a>active document</a> in the <a>top-level browsing context</a>
+        </li>
+        <li>
+          <var>settings</var> is not a <a>privileged context</a>
+        </li>
+        <li>
+          <var>environment</var> is not a <a>top-level browsing context</a>
```
mikewest: Drop this one, please. It's covered by the 'responsible document' check.

specs/credentialmanagement/index.src.html
```diff
@@ -747,20 +747,24 @@ <h3 id="request-credential">
...
     <li>
-      If <var>settings</var> does not have a <a>responsible document</a>, then
-      reject <var>promise</var> with a <code>SecurityError</code>
-      <strong>and</strong> terminate this algorithm.
-    </li>
-    <li>
-      If <var>settings</var>' <a>responsible document</a> is not the <a>active
-      document</a> in the <a>top-level browsing context</a>
-      If <var>environment</var> is not a <a>top-level browsing context</a>,
+      If any of the following conditions are met:
```
mikewest: Please restructure this as `Reject <var>promise</var> ... if any of the following conditions are met:`. IMO it reads more clearly if it's not broken up in the middle.

specs/credentialmanagement/index.src.html
mikewest: And, I know you just added in a previous patch, but please drop the `environment` definition as well. We don't need it, as the "incumbent settings object" is the settings object we're referring to with `environment`.

jmajnert:
Commit amended

mikewest:
Next time, please just add another commit rather than squashing and rebasing. That makes it easier to track changes. :)

Otherwise, LGTM. Thanks!

Issue #116, opened 2014-12-18
Explicitly mention the four cases around the HTTPS requirement in SRI


fmarier:
In terms of requiring secure origins or not, there are four cases:

1. http page loading an http resource
2. http page loading an https resource
3. https page loading an http resource (should be blocked by mixed content)
4. https page loading an https resource (not controversial, of course)

Whether or not we should allow the first two to have integrity attributes is still being discussed (for example [on the chromium security-dev list](https://groups.google.com/a/chromium.org/d/topic/security-dev/hJwYZOOPaH4/discussion)).

We should address these explicitly in the spec and tell implementers:
- whether they are allowed to ignore the integrity attribute in any of these cases
- whether they are allowed to block the sub-resource if it contains an integrity attribute

fmarier:
Actually, if we change at least one of the examples (as I've done in #181), it should be clear that SRI is orthogonal to TLS.

devd:
I am a bit uncomfortable doing that because I think (in 2015) all specs
should only use ssl in all their examples. But maybe thats just me?

metromoxie:
I'm with @devd. We should stick with HTTPS in our examples because that's how you should be writing code; everything else is a hack for backwards compatibility, as far as I'm concerned (okay, maybe that's a little harsh).

fmarier:
In that case, we'll need to write something about how user agents must honor the integrity attribute on both HTTP and HTTPS.

metromoxie:
Agreed. I'm happy to take this on and write up a pull request.

mozfreddyb:
Fixed by @metromoxie via #204 and #213.

PR #204, opened 2015-03-03
Remove refs to secure channel/insecure channel

### LINKED ISSUE #71: [SRI]: use the terms of (un)authenticated origins instead of (in)secure channels?
SRI currently uses and clarifies the terms *secure channel* and *insecure channel* in 2.1 (key concepts and terminology).

Do we instead prefer to talk about authenticated and unauthenticated origins, as we do in MIX?

devd:
We don't use these anyhow (per https://github.com/w3c/webappsec/issues/71#issuecomment-76702087)

fixes #71

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -144,15 +144,6 @@ This section defines several terms used throughout the document.
 The term <dfn>digest</dfn> refers to the base64-encoded result of
 executing a cryptographic hash function on an arbitrary block of data.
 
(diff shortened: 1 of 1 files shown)


fmarier:
Note: it's possible that @metromoxie will have to re-introduce a reference to that definition while fixing #116, but until then, we don't need it.

devd:
as Mike noted in the comment, we don't actually need the reference defined in the text if we are only using it in non-normative section, which is what @metromoxie  is planning to do iirc.

PR #205, opened 2015-03-04
No caching goal

devd:
We don't do this.


mozfreddyb:
`may enable improvements` sounded vague enough to leave in. But this change is alright to me :-)

PR #209, opened 2015-03-06
Clarify the conditionality of the upgrade mechanism

pde:
* The client must unfortunately send the "Prefer: return=secure-representation"
  header for all navigational requests, because HSTS will always need to be
  conditional on it.

* Clarify deployment advice with the concept of a "safely upgradable request",
  which is either a request to a mixed-content free origin, or a request from
  a client that supports the new upgrade mechanism.

* Refactor deployment advice generally.

### DIFF
--- a/specs/upgrade/index.src.html
+++ b/specs/upgrade/index.src.html
@@ -420,7 +420,15 @@ spec: RFC7240; urlPrefix: https://tools.ietf.org/html/rfc7240
   user-agent sniffing to make this decision, user agents MUST advertise their
   capabilities when making navigational requests by sending a
   <code>Prefer</code> HTTP request header expressing the desire for a secure
-  representation [[!RFC7240]] as described in [[#preference]].
+  representation [[!RFC7240]] as described in [[#preference]], unless they
+  have an HSTS directive active for the request's origin.
+
+  If a client has an HSTS state set for a given origin, the only reason to
+  send the <code>Prefer: tls</code> header is to trigger retransmission of the
+  HSTS header to extend its <code>max-age</code>.  Accordingly, clients SHOULD
+  send the <code>Prefer: tls</code> header if the <code>max-age</code> for an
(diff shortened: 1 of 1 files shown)


pde:
OK, nit fixed and definition moved to the "Key Concepts and Terminology" section.

specs/upgrade/index.src.html
```diff
@@ -307,6 +307,13 @@ <h2 id="key-concepts">Key Concepts and Terminology</h2>
   :: A {{Request}} is said to be <strong>upgraded</strong> if it is rewritten
      to contain a URL with a {{URL/scheme}} of <code>https</code> or
      <code>wss</code>.
+  :  <dfn export>safely upgradable requests</dfn>
+  ::  If an origin does not depend on this document's
+      <code><a>upgrade-insecure-requests</a></code> mechanism, all requests to
+      it are considered <strong>safely upgradable requests</strong>.  If an
+      origin depends on the upgrade-insecure mechanism, only requests from
+      clients that send a <a>Prefer</a>: <a>return=secure-representation</a> header are
```
pde: Technically requests from old clients that do not have mixed content blockers could also be considered "safely upgradable requests", but it's probably best to hope they go away rather than encouraging UA sniffing to detect them.

pde:
I've revised this pull request to at least almost-retire the header for HSTS sites.

pde:
The latest version of this PR has the nice side effect that it gives servers a neat way to send the HSTS header only very occasionally, reducing response sizes by 50-60 bytes.

mikewest:
I'm going to clean this up and merge it in, as we can't reasonably serve the "Downgrade unsupported browsers" use-case without it (or, at least, no one has made a counterproposal that is sane (mine were not)). I'm not happy about the impact, but given the requirement to take care of those sites that can't (or won't) update their content, this seems like the best we can do.

If and when we can come up with more substantial carveouts, we absolutely should.

PR #222, opened 2015-03-20
SRI: Use an equal sign in option-expression

Acubed:
Use of the equals sign has precedent in many protocols and syntaxes including HTTP, MIME, and HTML.
This production is forward-compatible with most all of them, e.g. quoted-string or urlencoded.

### DIFF
--- a/specs/subresourceintegrity/index.html
+++ b/specs/subresourceintegrity/index.html
@@ -143,7 +143,7 @@ substitute malicious content.</p>
 <code>script</code> element, like so:</p>
 
   <pre><code>&lt;script src="https://code.jquery.com/jquery-1.10.2.min.js"
-        integrity="type:application/javascript
+        integrity="type=application/javascript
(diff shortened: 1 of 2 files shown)


Acubed:
I need to fix examples, too.

devd:
again, I have publicly been in favor of this change so I shouldn't do the merging once you add the examples. 

@metromoxie @fmarier @mozfreddyb

fmarier:
I'm OK with merging this once the examples are fixed.

mozfreddyb:
:+1:

mozfreddyb:
(I find this slightly confusing to split on different characters for options and hash expressions though.)

Acubed:
@mozfreddyb It was like that before, right? (`hash-expression` being split with a `-`). And it's not really a key-value expression, anyways.

mozfreddyb:
Yeah, it's not really. Just slightly confusing :-)

metromoxie:
This lgtm, too. Merging.

Issue #217, opened 2015-03-18
SRI: Respond to mnot comments


devd:
see https://readable-email.org/list/public-webappsec/topic/sri-updates-to-the-spec-and-outstanding-issues#content-3

Opening this as a tracker task just to make sure it doesn't fall through the cracks.

mozfreddyb:
Re-reading to notice that this is some hefty feedback.
I'll try to break this down, but there are some items which I don't fully understand *yet*. Rigorous feedback from the other editors is highly appreciated here.

- [ ] 3.3.1 and elsewhere: usage of the term *resource*
- [ ] 3.2.2 definition of getPrioritizedHashFunction?
- [ ] RFC2119-happyness: use less MUST/SHOULD in uppercase if isn't conformance requirements
- [ ] 3.2.2 ambiguity between response and request headers
- [ ] 3.2.2. http authentication and integrity?
- [x] 3.6 integrity attribute ABNF (addressed in #214 and #220?)
- [ ] option-value chars should be
```
option-value-char  = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                                / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~" 
                                / DIGIT / ALPHA
                                ; any VCHAR, except delimiters
```

- [ ] 3.8. CSP directive: What if integrity-policy block occurs in CSP report-only header? Maybe let CSP header decide if report/block and just use `integrity`as a CSP directive
- [ ] 4. Proxies has clients add CC: no-transform in hopes that proxies won't modify the response body, but that's not how it works: <http://httpwg.github.io/specs/rfc7230.html#message.transformations> (no-transform only applies to the message it occurs within).
- [ ] Integrity data will be easier to add in some server-side scenarios if it's possible to convey it in HTTP headers on the response, rather than in the HTML (or CSS, or JS). For example 
```
HTTP/1.1 200 OK
Content-Type: text/html
Link: <https://widgets.example.com/foowidget>; rel="stylesheet"; type="text/css"; integrity="..."
```
- [ ] 3.2.1 agility still mentions ni URLs
- [ ] 3.3.3 talks about set of valued 'named information' links
- [ ] 1. introduction still talks about cache identifiers: "I think it's best to remove this, as it's speculative."


Also, CCing @mnot

mozfreddyb:
Another nit:
[ ] `parse metadata` does not talk about options yet.

devd:
[...]

PR #223, opened 2015-03-23
Fixing some easy nits from

mozfreddyb:
Fixing the following from issue #217:

* 3.2.2 definition of getPrioritizedHashFunction?
* 3.2.1 agility still mentions ni URLs
* 3.3.3 talks about set of valid 'named information' links
* 1. introduction still talks about cache identifiers: "I think it's best to remove this, as it's speculative."


mozfreddyb:
Note to self: Update .html file before merging

devd:
lgtm. I will let you merge since you want to also update the html.

fmarier:
Looks good to me too.

PR #232, opened 2015-03-30
Fixups

jmajnert:


jmajnert:
submit -> send already fixed. Will create another PR

Issue #235, opened 2015-04-02
SRI: One resource, many representations.


mikewest:
Given the prevalence of content negotiation schemes like https://www.igvita.com/2013/05/01/deploying-webp-via-accept-content-negotiation/, I worry a bit about the current MIME type syntax for integrity metadata. The previous syntax allowed each integrity metadata tuple to contain a distinct MIME type, which would, for instance, allow a server to return a WebP image rather than a PNG one for performance reasons if the UA supported it.

The current scheme has a single type, which applies to every algorithm/digest pair.

This might or might not be a reasonable limitation, but I'd like to see it explicitly considered, as I don't think I've seen any discussion of this impact.

/cc @igrigorik @yoavweiss @metromoxie @fmarier @mozfreddyb @devd

annevk:
We did: https://lists.w3.org/Archives/Public/public-webappsec/2015Jan/0288.html At the time I concluded it could always be added later, is that not true anymore?

mikewest:
I missed that thread, sorry. Thanks for dredging it up for me.

So the thought would be that `type:XXX`'s position determines which items it applies to?

yoavweiss:
It's fairly possible I'm missing something, but wouldn't the problem with conneg and SRI be that the hashes of the different representations would be different? Do we permit multiple hashes that some may not match the returned content? (I guess I could just go read the SRI processing model...)

mikewest:
Maybe _I_ should read the processing model. When I originally wrote the thing, I think this was in my head as something to support, but I haven't checked up on things to see where we've evolved. :)

As I noted, if you actively decide this isn't worth supporting, or is actively bad to support, great! Just note that explicitly in the spec somewhere.

annevk:
@mikewest yes. @yoavweiss you can already have multiple hashes due to different hashing algorithms.

yoavweiss:
@annevk - from [does resource match metadataList](https://w3c.github.io/webappsec/specs/subresourceintegrity/#does-varresourcevar-match-varmetadatalistvar) it seems like only the strongest hash is picked and validated against (so [there can only be one](https://www.youtube.com/watch?v=sqcLjcSloXs) ). I believe that if there are multiple "strongest" hashes, the first or last one (based on the UA specific priority function) will get picked.

annevk:
Right and if we ever wanted to support content negotiation you'd first filter on type and then on "strongest".

metromoxie:
[...]

PR #236, opened 2015-04-02
Support per-hash options and remove MIME type support

metromoxie:
This reworks grammar to support per-hash options. It also removes the mention of MIME types other than to mention possible support in the future. All of this is as discussed in w3c/webappsec#235.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -564,7 +564,7 @@ applied only to the `hash-expression` that immediately precedes it.
 
 <div class="note">
 At the moment, no `option-expression`s are defined. However, future versions of
-the spec make define options, such as [MIME types][].
+the spec make define options, such as MIME types [[!MIMETYPES]].
 </div>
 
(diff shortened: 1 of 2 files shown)


specs/subresourceintegrity/spec.markdown
```diff
@@ -571,6 +571,15 @@ value must be a valid [MIME type][].
 
 The `integrity` IDL attribute must [reflect][] the `integrity` content attribute.
 
+<div class="note">
+It should be noted that this syntax does not allow for [content negotiation][] if
+the developer specifies a MIME type. A change to the syntax to allow this may be
+considered in a future version of the spec, but for now, if a developer wants to
+use content negotiation, they will have to omit an `option-expression`.
```
fmarier: I would suggest being more explicit here and saying something like:

>  omit including a **type** option-expression.

metromoxie: Done.

fmarier:
LGTM

PR #237, opened 2015-04-03
"CSP"-Header ABNF adjusted

StefanOssendorf:
With the current ABNF the following header is possible:
CSP:     active    ,     active, active, active, active      

I think that's not intended.

### DIFF
--- a/specs/CSP2/index.html
+++ b/specs/CSP2/index.html
@@ -1079,7 +1079,7 @@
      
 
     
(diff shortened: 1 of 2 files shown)


mikewest:
Thanks!

annevk:
I'm not sure this is quite right. Don't we want to be able to signify multiple values going forward? If servers start doing an exact check for "active" that won't be possible.

PR #240, opened 2015-04-09
Use [SRI] as the list prefix

fmarier:
`[Integrity]` is a bit long and `[SRI]` seems to be the one that most people
have settled on despite the abstract suggesting `[Integrity]`.


mozfreddyb:
:+1:

devd:
thanks

PR #276, opened 2015-04-18
Removed references to shared cache (closes #207)

### LINKED ISSUE #207: Revise eligibility algorithm to require CORS
in *[Is resource eligible for integrity validation](http://www.w3.org/TR/SRI/#is-resource-eligible-for-integrity-validation)* the "shared" cache part needs removing

(see #190)

metromoxie:
This removes references to shared caches and publicly cacheable as it is a requirement that the resource be CORS enabled, as per #207.


specs/subresourceintegrity/index.html
```diff
@@ -146,7 +146,7 @@ <h2 id="introduction">Introduction</h2>
         integrity="sha256-C6CB9UYIS9UJeqinPHWTHVqh/E1uhG5Twh+Y5qFQmYg="&gt;
 </code></pre>
 
-  <p class="example highlight">Scripts, of course, are not the only resource type which would benefit
```
mozfreddyb: Are you not using the Makefile or do you have a different version of Kramdown? I'm on version 1.6.0, which keeps this attribute in place.

mozfreddyb: Oh, please excuse this worthless comment ;) It seems that this attribute is not associated with any extra styling anyway. Let's go ahead and merge!

PR #277, opened 2015-04-18
CREDENTIAL: Add 'Future Work by Other Groups'.

dlongley:
This PR adds a section to the spec about "Future Work by Other Groups", in particular the Web Payments IG and Credentials CG/potential WG. It spells out that those groups see this spec as useful and would like to extend it to support their own use cases. It also lists an issue about exploring the API and its data model further to ensure it can be easily and naturally extended in the future.

If this PR is accepted, then I think it's likely people from those groups would support the spec as a FPWD. The details over how to extend the API and any required further changes (if any) can be ironed out later.

The PR also removes some trailing whitespace :)


annevk:
I disagree with this. It may give the impression that WebAppSec won't extend this document.

dlongley:
> I disagree with this. It may give the impression that WebAppSec won't extend this document.

In what way? What language would you like changed to avoid that miscommunication?

mikewest:
I took a different approach to this in https://github.com/w3c/webappsec/commit/d275c5c988046f7fbafdeffcffef8f31b9984b8b (which you can more easily skim in https://w3c.github.io/webappsec/specs/credentialmanagement/#future-work-other-groups). I'd prefer to link off to the use case documents produced by other groups as informative references, as those are more likely to remain up to date than an explicit listing in this document.

WDYT, @dlongley, @msporny, @annevk, @hillbrad ?

(Thanks for pointing out the trailing whitespace! I killed those in https://github.com/w3c/webappsec/commit/91a589922433f97765b4b53270b8eec3734b6a95, just to keep editorial changes separated from cleanup. :) )

dlongley:
@mikewest, I think these changes look great! Thank you. If everyone else agrees, I think we can easily move forward.

hillbrad:
Is there any precedent for a Rec-track Technical Report referencing
speculative future work in this fashion?

References usually go the other direction the proposed future work (if and
when it is accepted on the Recommendation-track in a WG by the W3C
membership) would reference the baseline report it extends; the baseline
report doesn't begin by referencing speculative future work. References
exist to provide guidance on what is necessary to understand and implement
the specification.  They are not "shout outs".

I think there is plenty of time left for Recommendation-track technical
reports to mature out of these use-case documents, which may provide
concrete details of substantive technical value for implementers and users
of this report.  We can add such references, as appropriate, if and when
that happens.  I am not convinced that the use-case references add any such
value for the audience of this report at this time.

-Brad

msporny:
> Is there any precedent for a Rec-track Technical Report referencing speculative future work in this fashion?

yes, if you mark them up as &lt;p class="issue>. I think we need to at least make a note that there is work going on elsewhere that is attempting to align w/ the work being done in this group and vice-versa.

hillbrad:
OK, see:

https://github.com/w3c/webappsec/pull/282

dlongley:
Fine with me as an issue.

mikewest:
Merged Brad's patch, closing this one out.

Thanks folks!

akuckartz:
:+1: for finding a solution!

PR #283, opened 2015-04-20
Expand option-value-char characters

metromoxie:
This expands option-value-chars to a much wider array of characters as per @mnot's suggestion in #217. Note, however, that this also adds "/" as a character since we hope to support MIME type's in the future, and those distinctly use "/".

### DIFF
--- a/specs/subresourceintegrity/index.html
+++ b/specs/subresourceintegrity/index.html
@@ -695,7 +695,7 @@ option-expression  = option-name "=" option-value
 option-name        = 1*option-name-char
 option-name-char   = ALPHA / DIGIT / "-"
 option-value       = *option-value-char
-option-value-char  = ALPHA / DIGIT / "-" / "+" / "." / "/"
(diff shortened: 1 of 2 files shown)


mozfreddyb:
The Travis breakage seems to be strictly related to the credentials spec being broken. With all these specs going on, I wonder if we could automatically hide these warnings for unrelated pull requests. Or should we just split up specs into their own repo? cc @mikewest

mozfreddyb:
(This pull requests looks fine to me btw) :+1:

mikewest:
@tabatkins broke the Credentials spec, not me. https://github.com/tabatkins/bikeshed/pull/398 is out to fix it, which I assume he'll either pull or fix when he wakes up.

There aren't any Travis checks for SRI, no. If y'all want to go somewhere else, I'll be sad, but I'll deal. :)

mikewest:
Long story short, you can ignore Travis. I'd _love_ to be able to add some checks for you folks as well (Bikeshed catches lots of my dumb mistakes), but it's not clear whether or not ReSpec's warnings are easily surfaced...

mozfreddyb:
Thanks for the quick feedback! We'll go ahead with ignore-mode for a bit and bring up SRI testing in another thread (since we ruled out migrating to Bikeshed for v1). ;)

PR #292, opened 2015-04-22
CREDENTIAL: Generic matching algorithm

### LINKED ISSUE #274: CREDENTIAL: Use a generic matching algorithm that would work for custom Credentials types
The spec defines custom matching algorithms for the different sub-classes of `Credential`. If this could be restructured into a single-generic algorithm it could be used for custom sub-classes of `Credential`.

Possible use-case illustrated below but suggests that `options` could be elevated/merged to the parent as `types` is also just a filter OR `type` could be a property of `options`.

Matching algorithm would be something like:
*For each property of `options` eliminate all possible credentials where the `Credential` has a property with the same name and the value is not in the set of values provided.*

```javascript
//Store a FederatedCredential
navigator.credentials.store(
  new FederatedCredential({
    "id": "username",
    "federation": "https://federation.com"
  })
);

//Get a FederatedCredential filtered on the 'federation' property
navigator.credentials.get({
  "types": [ "FederatedCredential" ],
  "options": {
      "federation": [ "https://federation.com" ] //NOTE: This is no-longer called 'federations'
  }
});

// Assume we have defined an AgeVerificationCredential class which extends Credential
//  and defines it's type to be "ageVerification". There are some extra properties we might use
//  to do things like verification that are out of scope of this API

//Store a custom credential
navigator.credentials.store(
  new AgeVerificationCredential({
    "id": "person_12345_age",
    "isOlderThan18": true,
    "isOlderThan21": true,
    "issuer": "http://credentialissuer.com",
    "signature": { ... },
  })
);

//Get this custom credential filtered on some properties specific to this credential type
navigator.credentials.get({
  "types": [ "ageVerification" ],
  "options": {
      "isOlderThan18" : [ true ],
      "issuer" : [ "http://credentialissuer.com", "http://othertrustedissuer.com" ]
  }
});
```

adrianhopebailie:
This change updates the model for `Credential` to include an `isMatch()` function hook which is called during processing of the Credential request algorithm. The hook can be overridden by developers designing their own `Credential` classes in script therefor making the API flexible and extensible through ECMAScript.

Resolves #274


mikewest:
Before I skim in more detail, would you mind rebasing on top of master? This patch doesn't cleanly apply locally.

mikewest:
(Also, thanks for taking a stab at this!)

adrianhopebailie:
> Before I skim in more detail, would you mind rebasing on top of master? This patch doesn't cleanly apply locally.

@mikewest Not sure why, I mistakenly committed my .gitignore but have reverted that. The changes are only in index.src.html and were on top of the latest from master.

My git-fu is pretty basic, any suggestions on how to fix this?

mikewest:
```
git add remote w3c https://github.com/w3c/webappsec.git
git fetch w3c
git rebase w3c/master
[resolve conflicts]
```

Something like the above should do the trick! :)

adrianhopebailie:
Done. 

Mixing and matching betwen the GUI tool and command line is a mistake!
Already had w3c as a remote just hadn't done a rebase (thought Git for Mac did this automatically when you did a sync - clearly not).

This was my first adventure into writing WebIDL definitions so there may be some tweaks required however I think it get's the genera; gist across.

I'll add an example of using a custom credential (defined in script) later today as I think that will truly illustrate the reasons for the change.

specs/credentialmanagement/index.src.html
```diff
@@ -329,9 +331,16 @@ <h4 id="examples-post-signin">Post-sign-in Confirmation</h4>
         credential.send("https://example.com/sign-in/")
           .then(function (response) {
             if (response.status == 200)
+              // Inspect response to determine if this was a successful login then...
```
mikewest: Nit: Put this before the `if`, please. Otherwise you'll need `{}` around the clause for clarity (because it's multi-line).

specs/credentialmanagement/index.src.html
```diff
@@ -238,7 +238,7 @@ <h4 id="examples-password-signin">Password-based Sign-in</h4>
     <div class="example">
       <pre>
         navigator.<a attribute>credentials</a>.<a idl lt="get()">get</a>({
-          "<a idl>types</a>": [ "password" ]
+          "<a idl>filters</a>": { "password" : {} }
```
mikewest: I don't like this syntax for `password`. Passing in an empty object isn't at all intuitive, and since I see this as the core use case the current API is attempting to enable, I'd like to ensure that it is as streamlined as possible.

specs/credentialmanagement/index.src.html
```diff
@@ -495,16 +512,17 @@ <h4 id="interfaces-credential-types-credential"><code>Credential</code></h4>
     interface Credential {
       readonly attribute DOMString id;
       readonly attribute DOMString type;
+      boolean isMatch (any filter);
     };
     Credential implements Transferable;
   </pre>
   <dl dfn-for="Credential">
-    <dt><dfn attribute>id</dfn></dt>
+    <dt><dfn attribute for="Credential">id</dfn></dt>
```
mikewest: Nit: You don't need to add the `for` attributes; they're implicitly added by the `dfn-for` attribute on the containing `<dl>`.

specs/credentialmanagement/index.src.html
```diff
@@ -495,16 +512,17 @@ <h4 id="interfaces-credential-types-credential"><code>Credential</code></h4>
     interface Credential {
       readonly attribute DOMString id;
       readonly attribute DOMString type;
+      boolean isMatch (any filter);
```
mikewest: Nit: I'd prefer `matches`, but I guess it's still not clear to me what this is intended to model for credentials that aren't stored locally. That is, I understand the use case for sifting through locally stored passwords and federation data. I don't see how this "filtering" model fits with the kinds of cross-origin, third-party data that it seems it's meant to enable.

specs/credentialmanagement/index.src.html
```diff
@@ -515,13 +533,25 @@ <h4 id="interfaces-credential-types-credential"><code>Credential</code></h4>
...
       {{Credential/type}} attribute.
     </dd>
+    <dt><code><dfn method for="Credential">isMatch()</dfn></code></dt>
+    <dd>
+      {{Credential}} objects provide filter matching logic within the
+      <code>isMatch()</code> function. This hook allows different {{Credential}}
+      types to define how they will match the values from the {{CredentialRequestFilters}}
+      passed into the {{CredentialContainer/get()}} function. The base class implementation returns
+      <code>true</code>.
+    </dd>
```
mikewest: Nit: You can just add a newline here rather than a new `<dd>`. That will automagically create paragraphs, markdown-style.

specs/credentialmanagement/index.src.html
```diff
@@ -561,10 +591,10 @@ <h4 id="interfaces-credential-types-credential"><code>Credential</code></h4>
     </dd>
   </dl>
 
-  All {{LocallyStoredCredential}} objects MUST define an <dfn>options matching
-  algorithm</dfn> which returns <code>Match</code> if the object matches an
-  {{CredentialRequest/options}} dictionary, and <code>Does Not Match</code>
-  otherwise.
+  All {{LocallyStoredCredential}} objects MUST define a <dfn>filter matching
+  algorithm</dfn> in the body of the {{Credential/isMatch()}} function.
+  The algorithm must return <code>true</code> if there is a Match otherwise <code>false</code>
```
mikewest: Nit: s/Match/match/, s/Does Not Match/does not match/

specs/credentialmanagement/index.src.html
```diff
@@ -610,8 +640,8 @@ <h5 id="passwordcredential-matching">
     Matching Algorithm
   </h5>
 
-  {{PasswordCredential}} objects' <a>options matching algorithm</a> always
-  returns <code>Match</code>.
+  {{PasswordCredential}} objects' <a>filter matching algorithm</a> always
+  returns <code>true</code> (Match).
```
mikewest: Nit: s/ (Match)//

specs/credentialmanagement/index.src.html
adrianhopebailie: What do you suggest here? I'm looking for something that says "don't apply any filters".
How about

```javascript
navigator.credentials.get({
  filters : { "password" : "return all" }
```

Ultimately, it's not that important as the `PasswordCredential` is a special case as it isn't filtered. You either want them or you don't, you can't filter the set further.

specs/credentialmanagement/index.src.html
```diff
@@ -763,39 +804,50 @@ <h4 id="interfaces-request-options">
...
+    <dd>
+      For each possible credential that can be returned:
+      <ol>
+        <li>If the {{Credential/type}} is not a defined key then the {{Credential}}
+          is immediatley excluded.</li>
+        <li>If the {{Credential/type}} is a defined key then let V be the value in
+          the map stored under that key. Call {{Credential/isMatch()}} passing V
+          as the only argument. If the the function returns <code>false</code> then exclude
+          the {{Credential}}. </li>
+      </ol>
```
mikewest: Nit: This needs to be integrated into the request algorithm.

specs/credentialmanagement/index.src.html
```diff
@@ -763,39 +804,50 @@ <h4 id="interfaces-request-options">
...
+    <dd>
+      Each key in the map should correspond to the value of the
+      {{Credential/type}} property of any {{Credential}} type that
+      should be returned.
+    </dd>
+    <dd>
+      For each possible credential that can be returned:
+      <ol>
+        <li>If the {{Credential/type}} is not a defined key then the {{Credential}}
+          is immediatley excluded.</li>
```
mikewest: Nit: s/immediatley/immediately/

specs/credentialmanagement/index.src.html
```diff
@@ -763,39 +804,50 @@ <h4 id="interfaces-request-options">
...
   <pre class="idl">
-    dictionary FederatedCredentialRequestOptions {
-      sequence&lt;USVString&gt; providers;
-      sequence&lt;DOMString&gt; protocols;
+    interface CredentialFilterMap {
+      readonly attribute unsigned long size;
+
+      getter any get(DOMString credentialType);
+      setter creator void set(DOMString credentialType, any value);
+      deleter void remove(DOMString type);
```
mikewest: Why all this complexity? Why isn't a dictionary of dictionaries enough?

specs/credentialmanagement/index.src.html
adrianhopebailie: :+1:

specs/credentialmanagement/index.src.html
```diff
@@ -763,39 +804,50 @@ <h4 id="interfaces-request-options">
...
+      Each key in the map should correspond to the value of the
+      {{Credential/type}} property of any {{Credential}} type that
+      should be returned.
+    </dd>
+    <dd>
+      For each possible credential that can be returned:
+      <ol>
+        <li>If the {{Credential/type}} is not a defined key then the {{Credential}}
+          is immediatley excluded.</li>
+        <li>If the {{Credential/type}} is a defined key then let V be the value in
```
mikewest: Nit: s/V/<var>filter<\/var>/

specs/credentialmanagement/index.src.html
adrianhopebailie: I am learning WebIDL as I go here. My understanding was that dictionaries needed to define their keys up front which defeats the purpose of this whole proposal.

specs/credentialmanagement/index.src.html
```diff
@@ -763,39 +804,50 @@ <h4 id="interfaces-request-options">
...
+          the map stored under that key. Call {{Credential/isMatch()}} passing V
+          as the only argument. If the the function returns <code>false</code> then exclude
+          the {{Credential}}. </li>
+      </ol>
+      Built-in {{Credential}} types define an <a>filter matching algorithm</a> for that credential type.
```
mikewest: Nit: They define an implementation of `isMatch()`, right?

specs/credentialmanagement/index.src.html
```diff
@@ -821,9 +867,10 @@ <h4 id="interfaces-request-options">
 
     <pre>
       navigator.<a idl>credentials</a>.<a idl lt="get()">get</a>({
-        "<a idl>types</a>": [ "federated" ],
-        "<a idl>options</a>": {
-            "<a idl>providers</a>": [ "https://identity.example.com/" ]
+        "<a idl>filters</a>": {
+            "federated" : {
+                "<a idl>providers</a>": [ "https://identity.example.com/" ]
```
mikewest: Again, for the use cases I care about, it feels like this would make the API less intuitive. I think that's unfortunate.

specs/credentialmanagement/index.src.html
mikewest: Ah, I see, you've integrated it into the 'gather' algorithm below. Nevermind.

specs/credentialmanagement/index.src.html
adrianhopebailie: My intention is not about cro 

> I don't see how this "filtering" model fits with the kinds of cross-origin, third-party data that it seems it's meant to enable.

It is not intended to enable cross-origin third-party data. The intention is an API that allows for custom credentials defined in script (as opposed to requiring API updates).

specs/credentialmanagement/index.src.html
```diff
@@ -927,18 +975,6 @@ <h4 id="request-credential">
       </ol>
     </li>
     <li>
-      Let <var>type</var> be lowest common ancestor interface of the types
-      referenced in <var>request</var>'s {{CredentialRequest/types}} property.
-
-      ISSUE(w3c/webappsec#289): This is terribly hand-wavey. The intent is
-      clear, but I need to do the work to walk through the list of DOMStrings
-      and convert them to interfaces and etc. Busywork for later.
```
mikewest: I don't think removing this is possible quite yet, as we don't have a good mechanism by which to deal with user interaction sanely. Consider a new credential type you instantiate: since you'll be storing it in DOM storage, you won't need to ask the user for permission to grab any of the data (nor will it be protected from any script on the page grabbing it, but that's a distinct concern). You will, however, need to ask user permission for `PasswordCredential` objects, for instance.

Which triggers first? Do we walk through the password store first, triggering user mediation, and fall back to the new credential type iff the user doesn't grant permission? If a `PasswordCredential` is available without mediation, is it alright to simply return it, or do we need to walk through the new credential types as well and mix the data somehow to allow the user to choose a result?

The current spec avoids this by rejecting the promise if multiple kinds of credentials are requested (e.g. anything that isn't a `LocallyStoredCredential` (which perhaps needs to be renamed `CredentialStoreCredential` or something equally verbose). Dropping that constraint needs to be accompanied by a solution to the problem. I don't think we have it yet.

specs/credentialmanagement/index.src.html
```diff
@@ -998,8 +1032,8 @@ <h4 id="request-credential">
   {{LocallyStoredCredential}} isn't the only type? For example, how could an
   author pull some "RemoteCredential" that required interaction with a
   third-party, but fall back to a {{PasswordCredential}}? For the moment, we're
-  resolving this by rejecting the promise if the types listed in
-  {{CredentialRequest/types}} aren't all instances of the same type, but that
+  resolving this by rejecting the promise if the types listed as keys in
+  {{CredentialRequestFilters}} map aren't all instances of the same type, but that
```
mikewest: You dropped this constraint above. If you intend to drop it, we'll need to drop this issue as well. As I noted, it's not clear that we can.

specs/credentialmanagement/index.src.html
adrianhopebailie: Is your concern around losing the `types` property or the structure of the filters? To me this seems pretty easy to use (and understand) for anyone that just wants to use the built-in `Credential` types. Two examples explains it all.

For developers wanting to create custom types there is a lot more to wrap your ahead around but that should be expected.

mikewest:
Thanks for the PR! I've left a number of comments.

I'm fine with the central suggestion of exposing the matching algorithm, as that seems to allow some extensibility.

However, I don't like the fact that this makes the use cases the document actually intends to address more complex. Especially in the case of `PasswordCredentials`, the call is simply unintuitive, and will just add boilerplate to web developers' code. I'd also like to avoid the extra level of nesting for `FederatedCredentials`.

It's also not clear to me that `filter` is the right way to frame the thing that you're suggesting, as I don't understand how that verb makes sense for credential types that aren't stored locally in the browser's credential store. If I have to go out to a third-party site to grab data, I'm not filtering anymore. I'm doing something else. I couldn't come up with a term for that which made sense, and settled on the ultra-generic "options".

Does "filtering" fit into @dlongley's modeling of the world? It's not clear that it does.

specs/credentialmanagement/index.src.html
mikewest: It's important insofar as it's the use case I actively care about, and the use case I expect to see (in the short term) as 90% of the API's implementation in web developer's code.

If we keep the distinction between `types` and `filters`, we can deal with filterless types more cleanly. That is, the password case would remain `.get({ types: [ 'password' ] })`, as no filter would be required..

specs/credentialmanagement/index.src.html
mikewest: I'm misunderstanding the goal of custom credentials, then. I thought you wanted to subclass `Credential` in order to support things like `AgeVerificationCredential` (which I assumed would need to talk to some third-party in order to get data, via the `callback` POST that @dlongley talked about earlier). Is that not the case?

Can you help me understand how you'd intend to use this mechanism?

specs/credentialmanagement/index.src.html
mikewest: Sorry, that was confusing. I typed "dictionary" (which are indeed fixed sets of key/value pairs), I meant "object" (which is fluid and whatever).

specs/credentialmanagement/index.src.html
mikewest: This adds another layer of indirection. We've moved from:

```
.get({
  'federations': [ ... ]
})
```

to

```
.get({
  'options': {
    'providers': [ ... ]
  }
})
```

to

```
.get({
  'filters': {
    'federated': {
      'providers': [ ... ]
    }
  }
})
```

Maybe it's reasonable to accept that complexity if it actually solves use cases that folks care about. I feel less sanguine about accepting it when folks who care about those use cases don't think it's going to be enough.

mikewest: (Basically, I think I'd suck it up and deal with the extra layer if it seemed like that would get us to a compromise folks in the CCG and WPIG were happy with. But if those groups are unhappy, and planning another API anyway, then I'd prefer to drop the additional complexity we've added entirely rather than adding more. :) )

adrianhopebailie: Abstractly, if I have a function that queries a set of things and is expected return one of those things and I want to pass options to that function then an `options` object seems like the right pattern.

If that set contains things of various types and the query options must be type-specific then that options object needs a way of defining that. The only other approach I can think of is something like this (kind of a combination of our approaches) where anything at the top-level is a global option unless the property name is a type name, then it is a type-specific option:

```javascript
navigator.collection.query({
  "types" : ["thingType1", "thingType2", "thingType4"], //Filter the types I want
  "thingType1" : { "thingType1Option1" : true, "thingType1Option2" : 123 }, //Type-specific options for thingType1
  //No options needed for thingType2
  "thingType4" : ["a", "b", "f"], //Type-specific options for thingType1
  "globalOption1" : true,
  "gloablOption2" : function ( arg) { ... }
});
```

specs/credentialmanagement/index.src.html
adrianhopebailie: This is where my knowledge of the inner working of the browser are limited so I apologize in advance if I'm way off.

The way I envisioned this working was something like this:

1. The site owner provides the definition of the custom credential via a script.
1. The APi uses that definition to create a new instance of that credential (populated with the data stored in the credential store).
1. This credential is not returned to the caller without user mediation.

mikewest: 1. I can't speak for other browsers, but Chrome's credential manager sits on top of a relational database. Making it possible to store arbitrary data in the credential manager will be some work (and, given that we treat "credential" data separately from "browsing data", I worry a bit about the privacy implications of exposing an arbitrary data storage mechanism to the web that isn't easily cleared via the mechanisms users expect).

2. This still assumes that `LocallyStoredCredential` is what's being extended. It's not clear to me that something like `AgeVerificationCredential` is of the same kind. That is, the browser can't decide whether it has permission to grant access to the credential (or, really, access to the credential itself) without some sort of POST-based callback system that would require immediate user interaction. I don't have a good suggestion for merging that kind of model with this one. Do you?

specs/credentialmanagement/index.src.html
mikewest: I agree with the general suggestion that we need both a list of types we care about, and a list of options associated with each of those types. Perhaps a slight shift to place the two things in two separate arguments might make sense? That is:

Simple:
```
navigator.credentials.get(['password'])
```

Marginally complex:
```
navigator.credentials.get(
    ['password', 'federated'],
    {
        'federated': {
            'providers': [ ... ]
        }
    });
```

Complex:
```
navigator.credentials.get(
    ['password', 'federated', 'example'],
    {
        'federated': {
            'providers': [ ... ]
        },
        'example': {
            'property': true,
            'life': 43,
            'lifeOffset': 1
        }
    }
```

I don't really like this either, to be honest, but I don't have any better ideas.

It's not clear where `suppressUI` goes in this model, either. Perhaps another top-level `global options` argument? That gets ugly pretty quickly.

adrianhopebailie:
@mikewest I've been poring over this and I have come to conclusion that it won't work...

The issue is that trying to create something extensible to the developer means creating a callback that could potentially open a security hole since the code is defined and executed against the DOM.

A `Credential` type that allows something like this will need to be explicitly designed with this in mind and be able to handle the callback having potentially malicious code. Putting this in the base class is asking for trouble.

mikewest:
What's the threat model you're worried about? If an attacker can define an arbitrary callback, they can already execute arbitrary code in your origin, can't they?

adrianhopebailie:
Again, I may be showing my ignorance wrt the differentiation between the execution context of code behind a browser API vs that defined on the DOM within client scripts.

My proposal was for the site to provide a script that defines a custom `Credential` type and in so doing provide the execution logic for the `isMatch()` function. That means there is nothing stopping a rogue script from injecting logic that will be executed within what should be a secure context or am I wrong?

I like the idea of the extensibility you get from encapsulating the filtering logic in that function but don't really understand enough about the line between the two execution contexts within the browser to add anything more to this proposal.

specs/credentialmanagement/index.src.html
mikewest: I've taken a stab at this in https://github.com/w3c/webappsec/commit/e6c11ed448c9019620135227d750141408b207e8. WDYT?

mikewest: @adrianhopebailie ^^

adrianhopebailie: I certainly prefer it but as you say it may get unwieldy if the list of global arguments grows. Couldn't they also go in the `options` object? If the y doesn't match a type in the list of types it's assumed to be a global option?

mikewest:
This issue was moved to w3c/webappsec-credential-management#4

PR #293, opened 2015-04-23
CREDENTIAL: mismatch of "type" value between note and example

AxelNennker:
The text in 
http://w3c.github.io/webappsec/specs/credentialmanagement/#passwordcredential

[[type]]
    All PasswordCredential objects have their [[type]] slot’s value set to password.

while the example 
http://w3c.github.io/webappsec/specs/credentialmanagement/#examples-password-signin
has this code:
if (credential.type == "PasswordCredential") {

I think the text should be changed to:
[[type]]
    All PasswordCredential objects have their [[type]] slot’s value set to "PasswordCredential".


mikewest:
Good catch! That said, I'd suggest that the fix is the other way around. :) We're using shorter types in order to break the explicit link between types and interface names.

PR #298, opened 2015-04-24
SRI: Remove a redundant use case

fmarier:
It seems like the first use case (stylesheet) covers this pretty
well already.


mozfreddyb:
:+1:

PR #299, opened 2015-04-24
SRI: Remove references to unsupported elements

fmarier:
V1 only includes support for `script` and `link` so we should
avoid mentioning other types of elements.


mozfreddyb:
:+1:

PR #300, opened 2015-04-24
SRI: simplified error reporting

fmarier:
Now that we require same-origin or CORS, we can make these steps
simpler and assume that resources will be eligible.


mozfreddyb:
This accidentally contains the patch for #299, can you please rebase?

fmarier:
@mozfreddyb done!

PR #302, opened 2015-04-24
SRI: Add section Security Considerations: Behavior of chameleon resources

Acubed:
Remind people that a document, even with a successfully asserted hash, is still subject to multiple interpretations, including possibly malicious interpertations.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -748,6 +748,21 @@ will likely be difficult to avoid (image's `naturalHeight` and
 `naturalWidth` for instance).
 </section><!-- /Security::cross-origin -->
 
+<section>
+### Behavior of chameleon resources
+
+Resources may still exhibit differing behavior due to out-of-bound data
+like its media type, HTTP headers, or origin. Particularly, a "chameleon"
+resource is one that is valid under multiple media types. For example, a
+CSS document, validated as such to the server's satisfaction, might still be
+crafted with malicious effects that are triggered if interperted as
+application/javascript.
+
+Servers SHOULD NOT rely on "integrity" to ensure the safe behavior of
+user-generated content stored on untrusted/third party servers, even if
+the document is sanity/validity checked.
+</section><!-- /Security::behavior -->
+
 </section><!-- /Security -->
 
 <section>
--- a/specs/subresourceintegrity/index.html
+++ b/specs/subresourceintegrity/index.html
@@ -909,6 +909,22 @@ will likely be difficult to avoid (image’s <code>naturalHeight</code> and
   </section>
   <!-- /Security::cross-origin -->
 
+  <section>
+    <h3 id="behavior-of-chameleon-resources">Behavior of chameleon resources</h3>
+
+    <p>Resources may still exhibit differing behavior due to out-of-bound data
+like its media type, HTTP headers, or origin. Particularly, a “chameleon”
+resource is one that is valid under multiple media types. For example, a
+CSS document, validated as such to the server’s satisfaction, might still be
+crafted with malicious effects that are triggered if interperted as
+application/javascript.</p>
+
+    <p>Servers SHOULD NOT rely on “integrity” to ensure the safe behavior of
+user-generated content stored on untrusted/third party servers, even if
+the document is sanity/validity checked.</p>
+  </section>
+  <!-- /Security::behavior -->
+
 </section>
 <!-- /Security -->
 


Acubed:
I'd like to think I choose my terms carefully, but here I'm using "resource" "content" and "document" to mean essentially the same thing. What's the preferred term?

mozfreddyb:
Consider that the integrity attribute is tied to it's use case, i.e. the HTML tag. We do not have the common issue of images containing scripts, because why the would a web developer (or some sort of software) ever compute a hash and then include an image resource by using a script tag?

I am not opposed to your pull request, just a bit curious: How likely do you think is it that a web developer will hash something that is more than one file? For scripts I would expect this to be rather obvious.

Acubed:
Right, I'm not sure I can come up with any workable example that would actually be outright malicious at this time. But the functionality of "integrity" is liable to expand in the future, and therefore create more cases where people think "integrity" makes it safe to store documents on an untrusted server. The current draft specifically proposes adding integrity to CSS url() references.

This is mostly inspired by <https://bugzilla.mozilla.org/show_bug.cgi?id=470041>. Although caused by the embedding page changing the media type of the target page (against its advertised Content-Type), it was a result of parsing JSON as JavaScript.

The primary point is that "integrity" isn't an excuse to relax security standards, it supplements existing ones.

mozfreddyb:
Hey, could some other editor chime in?
This is an interesting contribution, given that there's a duplicate issue now ;)

devd:
I think this can be phrased more simply: something about how interpretation of a resource has other things not covered by the hash and SRI is intended only to prevent malicious modification of benign files and not for pinning behavior of malicious files.

Acubed:
That's pretty much what I'm saying in this PR, but with slightly more words.

(This is just a PR, so if multiple people are concerned about it, maybe keep an issue open too? I figure maybe someone has a better wording than mine with which to make a PR.)

specs/subresourceintegrity/spec.markdown
```diff
@@ -748,6 +748,21 @@ will likely be difficult to avoid (image's `naturalHeight` and
 `naturalWidth` for instance).
 </section><!-- /Security::cross-origin -->
 
+<section>
+### Behavior of chameleon resources
```
metromoxie: I've never heard the phrase "chameleon resources" before; is that a new or established term? I have a preference for "MIME type confusion," which is what I've always heard and used.

specs/subresourceintegrity/spec.markdown
```diff
@@ -748,6 +748,21 @@ will likely be difficult to avoid (image's `naturalHeight` and
...
+Resources may still exhibit differing behavior due to out-of-bound data
+like its media type, HTTP headers, or origin. Particularly, a "chameleon"
+resource is one that is valid under multiple media types. For example, a
+CSS document, validated as such to the server's satisfaction, might still be
+crafted with malicious effects that are triggered if interperted as
+application/javascript.
+
+Servers SHOULD NOT rely on "integrity" to ensure the safe behavior of
+user-generated content stored on untrusted/third party servers, even if
+the document is sanity/validity checked.
```
metromoxie: Maybe simply these paragraphs to something like:

"It is possible to generate documents that exhibit different but valid behaviors when presented with a different MIME type. An attacker can use this MIME type confusion to allow a document to pass an integrity check, but still have unexpected behavior. For this reason, servers SHOULD NOT rely on the integrity attribute to ensure safe behavior of user-generated content stored on untrusted servers."

Also, it might be worth mentioning that we may consider MIME types as an option to the integrity attribute in a future version of the spec, as we mention earlier.

metromoxie:
I like this, but like @devd, I think it can be simplified slightly. See my comments above.

specs/subresourceintegrity/spec.markdown
Acubed: I think I got the term from "Secure Content Sniffing for Web Browsers, or How to Stop Papers from Reviewing Themselves", Adam Barth et. el.: <http://www.adambarth.com/papers/2009/barth-caballero-song.pdf>

specs/subresourceintegrity/spec.markdown
Acubed: It's not just mime/media types, but maybe also things like CSP headers (in future editions) that enable/disable functionality (which should properly be in the media type, but whatever). An attack on changing the media type is the only one in the wild I'm aware of, though.

devd: I feel like Joel's text is saying exactly that. Would it help if he changed "Particularly" to "For example"?

Acubed: Change the word in my patch? Then there'd be two "For example"s in the paragraph and that sounds redundant said aloud

devd: Change the word in Joel's suggestion.

specs/subresourceintegrity/spec.markdown
devd: I think that paper was focused on sniffing and this section is more about a bunch of bytes having a completely different meaning based on mime confusion.

Although, I think a simpler title like "Impact of out-of-bound data" is much clearer and connects to the broader point we are making in the section.

specs/subresourceintegrity/spec.markdown
devd: whoops .. my mistake. I think I was referring to changing the word in your patch. WDYT?

specs/subresourceintegrity/spec.markdown
mozfreddyb: I've also heard the term 'Polyglots' in [Polyglots: Crossing Origins by Crossing Formats](http://www.cse.chalmers.se/~d02pulse/thesis/) by Jonas Magazinius et al., but I am not sure which is more common

devd:
@Acubed are you still working on this?

Acubed:
@devd Yeah, should I re-issue a pull request? I'm not sure I see a consensus here though, is there a specific patch to apply to this... patch?

devd:
me and @metromoxie both agreed on a simplification. If you are not opposed to the simplification, that would be the simplest.

fmarier:
Since we have moved to a new repository, please re-submit your updated PR on https://github.com/w3c/webappsec-subresource-integrity

PR #314, opened 2015-04-30
SRI: Allow non-eligible resources with empty integrity attributes

fmarier:
Using the original steps, a resource with an empty integrity
attribute might be blocked if it's not eligible (e.g. non-CORS and
not same-origin). This is wrong since we're not supposed to do
any checks or block anything in the "no metadata" case.

With this change, empty integrity attributes will always pass the
checks and never be blocked for SRI reasons.

### DIFF
--- a/specs/subresourceintegrity/index.html
+++ b/specs/subresourceintegrity/index.html
@@ -542,11 +542,11 @@ result is <var>newAlgorithm</var>, set <var>strongest</var> to
 
       <ol>
         <li>If <var>resource</var>’s URL’s scheme is <code>about</code>, return <code>true</code>.</li>
-        <li>If <a href="#is-resource-eligible-for-integrity-validation"><var>resource</var> is not eligible for integrity
-validation</a>, return <code>false</code>.</li>
         <li>Let <var>parsedMetadata</var> be the result of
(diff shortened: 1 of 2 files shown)


fmarier:
Here's an example of what would be blocked by the current algorithm:

    <script src="https://another-origin.example.com/foo.js" integrity=""></script>

(not eligible because it's non-CORS and not same-origin)

mozfreddyb:
Good catch!
Any other comments?

devd:
lgtm

metromoxie:
Unfortunately, this does not match Chrome's implementation. Why is this the right thing to do? It seems more logical to me to just not examine the integrity attribute at all if the resource isn't eligible.

metromoxie:
To be clear, though, it's not a big deal for us to change, I just want to make sure I fully understand why.

fmarier:
It seems to me that having a meaningful difference between

    <script src="https://elsewhere.com/foo.js">

and:

    <script src="https://elsewhere.com/foo.js"  integrity="">

reduces the usability of that feature because the first one is loaded and the second is blocked.

I would suggest that whether the integrity attribute is *missing* or *empty* should lead to the same outcome (loading the script).

metromoxie:
Okay then. I'll update Chrome.

PR #325, opened 2015-04-30
Grammar nit, remove false modesty.

hillbrad:


mikewest:
*shrug* I thought it was real modesty, but I wouldn't object to the current editors removing that bit. :)

hillbrad:
It's certainly not the case that none of it is new.  Acknowledging
influences is good, but there were a lot of proposals: nobody ever did all
the work all the way thorough the stack to make it consistent and secure
and privacy respecting.  There is plenty of new work here.  :)

fmarier:
Sounds good to me.

PR #336, opened 2015-05-06
POWER: Combine "Should X require priv. context" w/ threat model

diracdeltas:
Per discussion with @mikewest this morning, it seems that some of Mozilla's concerns raised in [this thread](https://lists.w3.org/Archives/Public/public-webappsec/2015Mar/0104.html) are addressed by reframing "Should X require a privileged context" in terms of the threat model.


mikewest:
LGTM. We'll probably tweak this some more going forward, but this is a good start. Thanks!

Issue #264, opened 2015-04-16
POWER: Privileged Contexts -> Secure Contexts


annevk:
Per @mikewest the rationale for the current name is OE. Given that OE is restricted to the `http` scheme it seems very clear that it is not HTTPS (Secure HTTP). Also, spelling privileged correctly is hard.

mikewest:
@bcrypt: Any objections, illustrious co-editor?

mikewest:
Er. @diracdeltas, apparently?

torgo:
Is there still a chance to influence this? IMO “secure contexts” is a kind of circular name for this document. The original title “powerful features” makes more sense: these are powerful features that require a secure context – the emphasis being on the features and what constitutes a powerful feature. With “secure contexts” I'm not sure what we're talking about – lots of things might require a secure context (e.g. passing a credit card number), but what the document is about is the requirement of secure contexts for powerful web features. Focusing on the features is what makes sense to me. If you don't like “powerful” how about “privacy-sensitive”? “Need for Secure Context for Privacy-Sensitive Features” maybe?

annevk:
Secure Context as in Secure HTTP but then for browsing contexts et al. Not sure how that does not make sense.

Powerful features does not work since we want to use this for non-powerful features too. And nobody can agree on what powerful features means anyway. So better to focus on the context and not the features.

torgo:
Sorry but I still don't get how it *does* make sense. When I try to describe it to developers, the "aha moment" is when I talk about features that are in some way more – won't use the p-word – dangerous / potential privacy infringing / intrusive / fraught. If we just say “secure contexts” then we could be talking about anything. The whole point here is that more of these kinds of features are making their way into the web and they should be limited to a TLS context. 

Sorry to be adding bike-shedding energy here but I do actually think this smatters because it frames the document and makes it clear why we're doing this. It also makes it more explainable.

diracdeltas:
I am okay with either name.

annevk:
What bit doesn't make use? There's browsing contexts, some are secure, some are not.

Given that Mozilla is considering restricting all new features to secure contexts, I don't think we can reasonably make a feature-based classification, and we need a context-based classification either way.

mikewest:
We went with "Secure Contexts".

PR #339, opened 2015-05-07
Please rename while we can

annevk:
The rationale for Privileged was apparently OE, but OE is HTTP-only and
therefore insecure. Given that, we should name this secure contexts
since it’s a much simpler name and easy to comprehend by everyone.

See also #264.

(I don't get the same references with Bikeshed. Not sure what is up with that. So you probably want to apply the patch locally.)

### DIFF
--- a/specs/powerfulfeatures/index.src.html
+++ b/specs/powerfulfeatures/index.src.html
@@ -1,4 +1,4 @@
-<h1>Privileged Contexts</h1>
+<h1>Secure Contexts</h1>
 <pre class="metadata">
 Status: ED
 ED: https://w3c.github.io/webappsec/specs/powerfulfeatures/
@@ -14,7 +14,7 @@ Abstract:
   authors for implementing features whose properties dictate that they be
   exposed to the web only within a trustworthy environment.
 Version History: https://github.com/w3c/webappsec/commits/master/specs/powerfulfeatures/index.src.html
-!Bug Reports: <a href="https://github.com/w3c/webappsec/issues/new?title=PRIVILEGE:%20">via the w3c/webappsec repository on GitHub</a>
+!Bug Reports: <a href="https://github.com/w3c/webappsec/issues/new?title=SECURE:%20">via the w3c/webappsec repository on GitHub</a>
 Indent: 2
 </pre>
 <pre class="anchors">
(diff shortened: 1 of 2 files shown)


specs/powerfulfeatures/index.src.html
```diff
@@ -191,12 +191,12 @@ <h4 id="threat-active">Active Network Attacker</h4>
   privacy who are able to target individual users, organizations or even
   entire populations.
 
-  <h3 id="threat-risks">Risks associated with non-privileged contexts</h3>
+  <h3 id="threat-risks">Risks associated with non-secure contexts</h3>
```
mikewest: Did you intentionally choose "non-secure" rather than "insecure"? I noted the same phrasing in https://blog.mozilla.org/security/2015/04/30/deprecating-non-secure-http/ and it sounds strange.

specs/powerfulfeatures/index.src.html
```diff
@@ -313,16 +313,16 @@ <h4 id="legacy-example">Example: Geolocation</h4>
...
 
-      If [[#settings-privileged]] returns <code>Not Privileged</code>, then
+      If [[#settings-secure]] returns <code>Not Secure</code>, then
       the algorithms should be aborted, and the <var>errorCallback</var> invoked
       with a <code>code</code> of <code>PERMISSION_DENIED</code>.
     </li>
     <li>
       User agents should announce clear intentions to disable the API for
-      unprivileged contexts on a specific date, and warn developers accordingly
+      unsecure contexts on a specific date, and warn developers accordingly
```
mikewest: s/unsecure/insecure/?

mikewest: Or non-secure, if there's a good reason for that term? Basically, we should be consistent.

annevk:
Done.

mikewest:
I'll give @diracdeltas a chance to veto, and land this tomorrow if she doesn't (hi, Yan!).

diracdeltas:
No veto from me.

Also, hi! I just discovered https://github.com/pulls so maybe I'll actually reply to issues I'm mentioned in. :)

mikewest:
Wunderbar.

annevk:
\o/

Issue #343, opened 2015-05-09
SRI: should we leave the options completely unspecified?


fmarier:
The per-hash option syntax is currently a bit weird because `?option1=value?option2=value2` almost looks like a query string, but not quite.

Since we're not defining any options in v1, I suggest we punt on this and simply say that anything after the question mark that's not whitespace is part of the option section. There's not much point in validating the unused option section at this stage.

In Firefox, we're parsing `sha256-AbcdE?options` like this:

1. look for a dash and extract `sha256` as the alg
2. look for a question mark and extract `AbcdE` as the digest
3. ignore everything after the question mark

mikewest:
+1. The option syntax seems strange, and the options-value-char list seems even more strange. If there aren't any options right now, then I'd suggest that there's no good reason to lock in a syntax for vague future things.

+@metromoxie

metromoxie:
sgtm. I'll update our implementation accordingly, and I'll upload a pull request to define this in the spec asap.

devd:
closeing per the merge above.

PR #344, opened 2015-05-11
Change option-expression to be *VCHAR

metromoxie:
As per the discussion in #343, we should unspecify option-expressions since no options are currently defined. This CL defines option-expression as simply *VCHAR to future-proof against any syntax we decide on.


devd:
lgtm

devd:
I am merging because I am certain all editors will agree: we can change things if anyone has concerns.

fmarier:
Thanks for that @metromoxie !

PR #348, opened 2015-05-12
Mixed content algorithms unaffected by SRI

metromoxie:
As mentioned in https://lists.w3.org/Archives/Public/public-webappsec/2015May/0065.html, we should probably have a note making our conversations about mixed content explicit. This adds a brief sentence stating that integrity doesn't affect the Mixed Content algorithms. /cc @devd @fmarier @mozfreddyb

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -679,8 +679,14 @@ document), just as they could alter the resource the hash is meant to validate.
 Thus, authors SHOULD deliver integrity metadata only to a [secure
 document][]. See also [securing the web][].
 
+Similarly, since integrity checks do not provide any privacy guarantees,
+[Integrity metadata][] MUST NOT affect the return values of the Mixed Content
+algorithms as defined in [section 5 of the Mixed Content][mixedcontent]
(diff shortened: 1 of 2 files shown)


mozfreddyb:
:+1:

specs/subresourceintegrity/spec.markdown
```diff
@@ -679,8 +679,14 @@ document), just as they could alter the resource the hash is meant to validate.
 Thus, authors SHOULD deliver integrity metadata only to a [secure
 document][]. See also [securing the web][].
 
+Similarly, since integrity checks do not provide any privacy guarantees,
+[Integrity metadata][] MUST NOT affect the return values of the Mixed Content
+algorithms as defined in [section 5 of the Mixed Content][mixedcontent]
```
gerv: defined in _the_ ?

devd:
Nit: I agree with the diff, but I find it very weird to use MUST NOT for a completely different spec. To me, it is already obvious based on this spec and this clarification is just in case there is misunderstanding. DOES NOT seems more right to me.

devd:
lgtm

PR #352, opened 2015-05-17
Remove UI integrity from UA use case

### LINKED ISSUE #321: [SRI] Remove browser UI integrity use case?
Section 1.2.1 lists a user agent which composes its UI from HTML as a potential use case for SRI.  I'm not sure this is a good example, first because I struggle to imagine a set of trust boundaries which would allow a malicious party to modify files describing trusted UI but not also be able to subvert the integrity mechanism.  Also, since file schemes are pretty much excluded, this makes it even more difficult to conceptualize how this would be implemented.  I think we are better off leaving this out than having a potentially confusing example.

devd:
Fixes #321


mozfreddyb:
:+1:

PR #353, opened 2015-05-17
SRI: sort authors alphabetically by first name

fmarier:
Swap Dev and Freddy to make the order alphabetical.


devd:
I am willing to accept and merge, but it seems oddly self-serving so I will let one of the other editors do it.

PR #354, opened 2015-05-17
alert example should escape quotes

devd:
This will mean that kramdown doesn't fancy it up and keeps it normal.


mozfreddyb:
Isn't this a use case for backticks?

devd:
this is more about the markdown feature that makes quotes (and probably backticks) more fancy and friendly to view. We can switch to backticks if you prefer, but I imagine that will stick require escaping with a backslash.

PR #355, opened 2015-05-17
Recalculate hashes in agility section

devd:
We updated the example text and forgot to update the hashes here.


specs/subresourceintegrity/spec.markdown
```diff
@@ -247,15 +247,15 @@ resource in order to provide agility in the face of future discoveries.
 For example, the "Hello, world." resource described above may be described
```
fmarier: Now that the resource is no longer "Hello, world.", perhaps we should rephrase that and just refer to "the resource described in the previous section may be described..."

devd:
@fmarier wdyt?

fmarier:
Looks good

fmarier:
I just rebased and merged this on master.

PR #356, opened 2015-05-17
Point to right chunk of CSP ABNF for hash syntax

devd:
In https://w3c.github.io/webappsec/specs/content-security-policy/ hash-src is defined with quotes and we don't use that.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -205,7 +205,7 @@ At the moment, no options are defined. However, future versions of
 the spec may define options, such as MIME types [[!MIMETYPE]].
 </div>
 
-This metadata MUST be encoded in the same format as the `hash-source`
+This metadata MUST be encoded in the same format as the `hash-algo`
 in [section 4.2 of the Content Security Policy Level 2 specification][csp2-section42].
 
 For example, given a script resource containing only the string "alert('Hello, world.');",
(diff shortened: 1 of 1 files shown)


manger:
hash-algo is only how to identify the algorithm ("sha256" / "sha384" / "sha512"). It doesn't give the syntax for the hash value.
A better fix would be:

-This metadata MUST be encoded in the same format as the `hash-source`
-in [section 4.2 of the Content Security Policy Level 2 specification][csp2-section42].
+This metadata MUST be encoded as per `hash-with-options` in section 3.6 "The `integrity` attribute".

specs/subresourceintegrity/spec.markdown
```diff
@@ -205,7 +205,7 @@ At the moment, no options are defined. However, future versions of
 the spec may define options, such as MIME types [[!MIMETYPE]].
 </div>
 
-This metadata MUST be encoded in the same format as the `hash-source`
```
fmarier: How about "as the \`hash-source\` (without the single quotes)"?

devd: that sounds good to me.

devd:
@fmarier @manger wdyt?

fmarier:
Sorry, what I meant was literally this string:

    ... same format as the `hash-source` (without the single quotes) in ...

devd:
hah ok

manger:
It would be better to refer to the syntax definition later in the same doc, instead of referring to the separate CSP doc in a slightly different way than that definition. But it is ok.

PR #358, opened 2015-05-18
Use domains reserved for examples, complete examples, and consistent …

manger:
Use example.com and example.net, instead of domain names that might be real. [see RFC 6761 Special-Use Domain Names]
Use full hashes, instead of eliding part of the hash.
Only use the same hash where it could make sense (not for style.css and include.js).


devd:
thanks!

PR #368, opened 2015-05-18
CSP: fix broken hash value in 7.15.2

michaelficarra:
Two issues here: the hash algorithm should be sha256, and the hash itself should be base64 encoded. The current example uses a base64 encoding of the hex encoding of the hash.


mikewest:
LGTM; I didn't check the hash, but I assume you did!  :)

michaelficarra:
I checked it, don't worry.

PR #369, opened 2015-05-18
getprioritizedhashfunction is not exposed to webapps

### LINKED ISSUE #346: [SRI] getPrioritizedHashFunction should be a normative definition
Section 3.2.2 states that a user agent MUST provide a means of determining algorithm priority, but doesn't normatively specify an API, only gives the example of "getPrioritizedHashFunction(a, b)".  If we are going to be able to test this requirement as a MUST, it needs to be normatively defined with a way to access it.

devd:
fixes #346

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -285,6 +285,9 @@ collision-resistant.  For example, `getPrioritizedHashFunction('sha256',
 'sha512')` would return `'sha512'` and `getPrioritizedHashFunction('sha256',
 'sha256')` would return the empty string.
 
+The `getPrioritizedHashFunction` implementation is vendor specific, and the interface 
+provided is not exposed to the application.
+
 </section><!-- /Framework::Cryptographic hash functions::Priority -->
 
 </section><!-- /Framework::Cryptographic hash functions -->


specs/subresourceintegrity/spec.markdown
```diff
@@ -285,6 +285,9 @@ collision-resistant.  For example, `getPrioritizedHashFunction('sha256',
 'sha512')` would return `'sha512'` and `getPrioritizedHashFunction('sha256',
 'sha256')` would return the empty string.
 
+The `getPrioritizedHashFunction` implementation is vendor specific, and the interface
```
fmarier: The verb tense here doesn't feel quite right. The previous paragraph says "if a user agent implemented getPrioritizedHashFunction() then it would ..." and in this change, it feels like we assume that it exists.

I suggest rewording to something like "Note that getPrioritizedHashFunction() is never exposed to applications given that both its implementation and its existence are vendor-specific."

mozfreddyb:
Oops, my merge mid-aired with @fmarier's comment. Sorry.

mozfreddyb:
I have reverted this in #377 and was hoping this would allow me to reopen here, but it doesnt :disappointed:

PR #371, opened 2015-05-18
SRI: update hash deprecation wording

devd:


specs/subresourceintegrity/spec.markdown
```diff
@@ -263,9 +263,9 @@ list, and use that metadata to validate the resource (as described below in
 the "[parse metadata][parse]" and "[get the strongest metadata from
 set][get-the-strongest]" algorithms).
 
-When a hash function is determined to be insecure, user agents MUST deprecate
-and eventually remove support for integrity validation using that hash
-function.
+User agents MAY deprecate support (by blocking loads) for integrity validation using
```
fmarier: I'm ok with this, but we should probably mention this on the list first since it is a departure from our otherwise "fail open" policy.

devd: Well, my main concern with fail open is that we shouldn't limit our options for a future new version of SRI, so this isn't quite that. But, yeah, you have already responded on that thread. I mostly think we have reached the point of diminishing returns on what we should do here, so I am happy with whatever the consensus is ;)

fmarier:
Just noting that this was brought up by @gerv on https://lists.w3.org/Archives/Public/public-webappsec/2015May/0070.html

fmarier:
I've just put an alternative on #383.

specs/subresourceintegrity/spec.markdown
fmarier: Wouldn't a user agent which has deprecated and/or removed `sha256` just load the subresource without performing any integrity checks? It would be like the case where the only hash algorithm in the integrity metadata is an unknown one (e.g. `md5`).

gerv:
I can't see a situation where a UA would ever want to block loads, given that the no-integrity default is to allow loads. Given that 99.999% of users will not be under attack, and that the user can do nothing about the site's hash algorithm selection, that would be a "break the site for no benefit" move.

devd:
I am not sure if I am creating a red herring or not, but isn't there the (happening right now) fact that we deprecate SHA-1 for SSL; when the no SSL option is to load HTTP?

fmarier:
@devd You're right that this is an example of failing closed and it's quite possible we'll take the same approach when we have to deprecate SRI hash functions.

I just don't think we should enshrine a particular tactic (for lack of a better word) in the spec. Like you said [on the list](https://lists.w3.org/Archives/Public/public-webappsec/2015May/0087.html) deprecating broken algorithms is an art and a long game :)

devd:
Sorry, if it wasn't clear: I agree with the wording of "MAY" or "SHOULD". I thought Gerv was not convinced with that wording, which is what I was pushing back against.

fmarier:
@metromoxie just merged #383.

PR #383, opened 2015-05-25
SRI: give more flexibility to user agents during hash deprecation period

fmarier:
This was suggested by @gerv on https://lists.w3.org/Archives/Public/public-webappsec/2015May/0070.html

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -263,8 +263,9 @@ list, and use that metadata to validate the resource (as described below in
 the "[parse metadata][parse]" and "[get the strongest metadata from
 set][get-the-strongest]" algorithms).
 
-When a hash function is determined to be insecure, user agents MUST deprecate
+When a hash function is determined to be insecure, user agents SHOULD deprecate
 and eventually remove support for integrity validation using that hash
+function. User agents MAY check the validity of hashes which use a deprecated
 function.
 
 To allow authors to switch to stronger hash functions without being held back by older


fmarier:
This is an alternative to #371. The difference:
- it maintains the requirement that user agents deprecate (and eventually remove) broken algorithms
- it give flexibility to user agents during the deprecation period (they can choose to ignore or honor the integrity attribute)

Since deprecation of security features is always tricky and involves juggling a number of trade-offs, I would vote for not being too strict here so that we can work out a spec-compliant solution when the time comes.

On the other hand, I think that the actual deprecation should not be optional. If a hash function is deemed weak or broken, we should at the very least spam the devtool console.

devd:
What do you think of my suggestion in https://lists.w3.org/Archives/Public/public-webappsec/2015May/0087.html (and why I don't think we should say MUST for the reasons outlined)

fmarier:
> One thing I worry about is that the MUST wording might limit UAs---if a UA wants to deprecate a hash function in one corner of its implementation, the current wording might force its hand for SRI (and thus affect the ability to deprecate an old hash function in another corner).

I read the MUST in the original wording as "user agents MUST do something about deprecating this algorithm". That something could be "showing a warning in the console".

> Additionally, deprecation is an art and a long game. For example, UAs might actually go out and block known broken hashes to make sure the world moves on to better hashes or it may decide to only warn and then start blocking a year later. Or do something more weird involving the set of permissions granted to the script etc.

I think all of the above are compliant with the requirement to do *something* about the broken algorithm. Am I reading this too liberally?

gerv:
If you are going to say that user agents MUST deprecate a hash function which is "determined to be insecure", you will need to be more specific about who does the determining and how that's officially decided. 

I think that instead, you should make this a SHOULD, because that shows the spec authors think deprecation is a good idea without running into the definitional problems I just outlined.

fmarier:
Fair enough. You both make a good case that MUST is perhaps too strong here.

I've changed that to a SHOULD.

mozfreddyb:
I have no objections to this change. I think @metromoxie was very active in that discussion. Flagging him.

metromoxie:
looks fine to me. merging.

Issue #238, opened 2015-04-05
SRI: Integration with Fetch


annevk:
I think point 5 is missing that we basically don't want to give the page any kind of notification until we have end-of-file, no? So any callbacks need to be suppressed until we can verify the integrity and once that is done we can run all three in sequence.

Also, rather than having section 3.9, that should also be a modification to Fetch I think. We should simply return a network error and that will make HTML do the right thing. And since only `script` and `link rel=stylesheet` can set integrity to begin with, not much else is needed.

metromoxie:
@annevk:

To your first point, perhaps I'm misunderstanding, but it seems like it's not that we fundamentally don't want notification until we have an end-of-file. Isn't it okay to get a separate network error earlier, since that's equivalent? What 'notification' are you concerned about?

To your second point, I agree that it would be better to rewrite 3.9 in terms of Fetch. I'm trying to figure out how best to do that. Should it be as simply as something like adding a modification before step 14 in https://fetch.spec.whatwg.org/#process-response-end-of-file such as:
  "If the response's integrity state is corrupt, set response to a network error."

metromoxie:
Also, @mozfreddyb and @devd, you should definitely take a look at this thread as well. We should deal with this before going much further (and @fmarier, but you're already CC'd).

annevk:
@metromoxie in the future `<link>` and `<script>` might expose the underlying response stream. The way that stream gets fed is through those "process response body" callbacks.

I think before step 15 (note that 14 is about the request), you need to check if request has a non-null integrity field (you'll need to add this field to request as well). Then if it does, you need to run substeps where you wait for the entire body to be pushed. Then you do the integrity check using request's integrity field. And if that passes you queue the three callbacks (response, response body, response end-of-file).

Alternatively around step 10 you do a similar check. If integrity is set you simply wait until the body reaches end-of-file and do the check then and modify response to a network error as appropriate.

The second one might be a little nicer since that's where Mixed Content, nosniff, and soon CSP also happen.

metromoxie:
Thanks, that helps a bunch, and the alternative you suggest at step 10 makes particularly a lot of sense. A quick clarification though. In 3.4 step 1, we already add "integrity metadata" to the request. This should be sufficient, right? That is, we don't need to add an additional "integrity field," that you referred to, correct?

metromoxie:
[...]

PR #390, opened 2015-05-27
SRI: stop whitelisting the `about` scheme (fix #319)

### LINKED ISSUE #319: SRI: Should the about: scheme be whitelisted?
In #280, @mozfreddyb asks: "should we really mandate that the about: scheme has always good integrity? It may be current practice, but I don't think we should rely on about things being from-disk content until eternity."

fmarier:
According to @mikewest, this was put in the spec to deal with
iframes. Since v1 doesn't include support for that element, we
can remove it from the spec.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -416,15 +416,14 @@ the user agent.
 <section>
 #### Does <var>resource</var> match <var>metadataList</var>?
 
-1.  If <var>resource</var>'s URL's scheme is `about`, return `true`.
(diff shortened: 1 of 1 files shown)


mozfreddyb:
Note that this pull request just changes the markdown file, not the HTML

fmarier:
Yeah we've got a few outstanding pull requests, I'll update the markdown and gh-pages once we've merged a couple.

mozfreddyb:
Yeah, alright.
(I think I ought to re-publish on w3.org as soon as we got fetch integration ( #238))

devd:
lgtm

PR #391, opened 2015-05-29
SRI: Simplify the element modifications to just set integrity metadata

metromoxie:
This modifies the algorithms to more closely integrate with Fetch, as discussed in #238. This makes several relevant changes:
   1. Adds modifications to main Fetch, after the Mixed Content check, to check for valid integrity. If it fails, it returns a network error.
   2. It removes the "integrity state" on the response, as this state doesn't need to be passed around anymore (everything is done in main fetch). This removes a bunch of text from various algorithms.
   3. It simplifies the element modifications to just set `integrity metadata` on the request before Fetch occurs.


metromoxie:
Latest version is "feature complete." Let me know how it looks.

devd:
nice!

lgtm

(only read the markdown)

specs/subresourceintegrity/spec.markdown
```diff
@@ -595,15 +577,15 @@ failed resource with a different one.
...
 Whenever a user agent attempts to [obtain a resource][] pointed to by a
 `link` element that has a `rel` attribute with the value of `stylesheet`,
-insert the following step between steps 1 and 2:
+modify step 4 to read:
 
-1.  Set the [integrity metadata][] of the request to the value
-    of the element's `integrity` attribute.
-{:start="2"}
+Do a potentially CORS-enabled fetch of the resulting absolute URL, with the
+mode being the current state of the element's crossorigin content attribute,
```
mozfreddyb: I'm not sure I parse this correctly.
Would this send credentials if the crossorigin attribute is "use-credentials"?

In any case, we still rely on CORS to cooperate, don't we?

devd: yes, it would, as I read it.

metromoxie: As a clarification, this is an exact copy of the text already present in http://www.w3.org/TR/html5/document-metadata.html#concept-link-obtain, just extended with ", and the [integrity metadata][] of the request to the
+value of the element's `integrity` attribute", so really the only part of this up for debate is that last clause :-)

And, yes, this still requires CORS because the "match" algorithm still does the "Is resource eligible for integrity validation check," which will return "false" if resource is not same origin or CORS enabled.

metromoxie:
I spoke with @annevk, and he is unavailable to review this for a few weeks. He suggested that if the other editors approve, we should just commit it, and he can review and we can amend at a later point. Thus, if @fmarier and @mozfreddyb approve, we should merge this whenever we're read.

mozfreddyb:
@fmarier said he'll look when he's back from the conf (or maybe while at the conference).
Let's wait for him to merge then :)

specs/subresourceintegrity/spec.markdown
```diff
@@ -472,49 +472,32 @@ to enable the rest of this specification's work [[!FETCH]]:
...
-        the <var>response</var>'s integrity state to `indeterminate` and
-        skip directly to firing the event.
-
-    2.  If <var>response</var> [matches][match] the request's integrity
-        metadata, set the <var>response</var>'s integrity state to `intact`
-        and skip directly to firing the event.
-
-    3.  Set the <var>response</var>'s integrity state to `corrupt`
-        and skip directly to firing the event.
+    1.  If <var>request</var>'s integrity metadata is not the empty string,
```
fmarier: I would suggest the same phrase as the one you used above: "is a non-empty string"

Because I consider `integrity=""` and `integrity="         "` to both be empty strings for the purposes of SRI.

specs/subresourceintegrity/spec.markdown
```diff
@@ -610,15 +594,15 @@ failed resource with a different one.
...
+Do a potentially CORS-enabled fetch of the resulting absolute URL, with the
+mode being the current state of the element's crossorigin content attribute,
+the origin being the origin of the link element's Document, the default origin
+behaviour set to taint, and the [integrity metadata][] of the request to the
+value of the element's `integrity` attribute.
 
-Additionally, in the paragraph specifying when to fire a `load` event, add
-"response with a `corrupt` integrity state" to the list of load failure
-reasons which are considered network errors.
+{:start="2"}
```
fmarier: Does this need to be updated to "4" now that this is replacing step 4?

specs/subresourceintegrity/spec.markdown
metromoxie: Done.

specs/subresourceintegrity/spec.markdown
metromoxie: TBH, I don't know what the {:start="2"} business does, and it doesn't modify the compiled index.html at all :-) Thus, I happily have changed it to "4".

fmarier:
I'm certainly not going to claim I'm an expert on Fetch, but this looks fine to me. Thanks @metromoxie !

metromoxie:
Okay, I'm going to merge, and then we can get comments from @annevk whenever he is available again. Thanks!

Issue #317, opened 2015-04-30
SRI: Clarify how to handle non-eligible and invalid metadata


fmarier:
I think we need to do a follow-up to  #314 but I'd like us to agree on the expected behavior before I do that.

What I'm interested in is how to deal with non-eligible resources that have empty, valid or invalid metadata, as well as eligible resources with invalid metadata.

Let's consider these non-CORS examples:

1. Non-eligible, empty metadata: `<script src="http://elsewhere.example.com/foo.js" integrity="">`
2. Non-eligible, valid metadata but unsupported algorithm: `<script src="http://elsewhere.example.com/foo.js" integrity="rot13-base64hash">`
3. Non-eligible, invalid metadata: `<script src="http://elsewhere.example.com/foo.js" integrity="invalid">`

Here, I would suggest that the first one should be loaded because it's equivalent to the pre-SRI web. The presence of an empty `integrity` attribute shouldn't change anything for authors. This is what I fixed in #314.

The last two on the other hand, should probably be blocked because they represent an attempt by the author to use SRI badly and by refusing to load their resources we ensure that they notice immediately and fix the problem. Any objections?

Now, let's look at the eligible cases (same-origin):

1. Eligible, empty metadata: `<script src="foo.js" integrity="    ">`
2. Eligible, valid metadata but unsupported algorithm: `<script src="foo.js" integrity="rot13-base64hash">`
3. Eligible, invalid metadata: `<script src="foo.js" integrity="invalid">`

Here, I would suggest that the first two be loaded. The first example is obvious since everything is fine. In the second case, we already decided to "fail open" to be forward compatible with future hash algorithms.

The third case is different because we now have a mechanism to add arbitrary per-hash options. We could decide that this is how future versions of the spec will need to extend the syntax and block subresources that consist only of invalid metadata (as opposed to valid but unsupported hash algorithm). This would allow authors to easily catch mistakes like `integrity="sha256bash64hash"` (i.e. forgetting the dash). Or we could fail open in this case too. Thoughts?

fmarier:
Just to comment regarding "implementationability", what I've suggested above is what I've implemented so far in Firefox.

devd:
I think third should fail open too. I am generally against UAs trying to fail closed: I worry about introducing a new syntax in the future that will then break old browsers.

Also, this would be a great test case *hint* *hint* @hillbrad ;)

fmarier:
[...]

PR #394, opened 2015-06-05
SRI: Fail open on ineligible

metromoxie:
As per the discussion in #317, I believe that ineligible resources should fail open, not closed. More specifically, since a CORS failure because of a lack of CORS headers from the server will result in a network error (and thus SRI will never be activated), this has the practical effect that if a developer forgets to add the crossorigin attribute, the integrity should not be enforced. Thoughts @devd @mozfreddyb @fmarier?


fmarier:
Now that eligibility is down to just checking for CORS or same-origin, I think this makes sense.

metromoxie:
Okay, I'm going to merge, since @fmarier and @devd seem to agree. @mozfreddyb, feel free to jump in if you have a different perspective.

PR #403, opened 2015-06-16
SRI: no need to add "Cache-Control: no-transform"

fmarier:
This directive only applies to the request it's in. It doesn't
apply to the response, which was the whole point of us adding that
header.

https://github.com/w3c/webappsec/issues/217#issuecomment-112223659


annevk:
I guess you might still want to encourage developers to include this header on the response?

fmarier:
Good point, I will revise this.

fmarier:
I've added a blurb about including the header in responses, as suggested by @annevk.

PR #405, opened 2015-06-16
remove cache-control bits

devd:
Per discussion in #217


devd:
lol .. missed the PR from @fmarier

PR #407, opened 2015-06-19
Lowercase 'must' regarding priority of hash functions

metromoxie:
As per an offline conversation, it feels odd to upper-case 'MUST' this statement since it's effectively an implementation detail, and not related to anything externally facing. Thus, it makes more sense to lowercase this. Confirmation from @devd, @fmarier, or @mozfreddyb?

### DIFF
--- a/specs/subresourceintegrity/index.html
+++ b/specs/subresourceintegrity/index.html
@@ -399,7 +399,7 @@ stronger hash functions as they become available.</p>
     <section>
       <h4 id="priority">Priority</h4>
 
-      <p>User agents MUST provide a mechanism of determining the relative priority of two
+      <p>User agents must provide a mechanism of determining the relative priority of two
 hash functions and return the empty string if the priority is equal. That is, if
 a user agent implemented a function like <dfn>getPrioritizedHashFunction(a,
 b)</dfn> it would return the hash function the user agent considers the most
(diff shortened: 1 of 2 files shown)


fmarier:
Looks good. It was weird to have this written as a strong requirement.

annevk:
This is not have normative language works. If this is not a requirement, you should not use must, lowercase or uppercase.

metromoxie:
Re-reviewing this, I hate to be that guy, but I actually like the original normative wording, so per @annevk's point, I think we should restore the capital MUST. It *is* a user agent requirement, because it is per-user agent, and it is used by the algorithms. I think our note is sufficient for pointing out that it's an implementation detail.

@fmarier @mozfreddyb @devd, what do you think? Should we revert my change?

fmarier:
I'm ok with removing the word must altogether as per @annevk's comment.

Neither the function nor the algorithms have to be implemented in the user agent. They're just there to illustrate how user agents should parse and verify integrity attributes.

annevk:
Returning the empty string does seem like a normative requirement, but I might not have enough context.

fmarier:
The part that is normative is the prioritization of algorithm strength by user agents and the grouping of hash digests with the same hash algorithm. Whether the mechanism that the user agent implements returns a string or a zero, or anything else, doesn't actually matter.

metromoxie:
So @fmarier, you want to keep the paragraph, but remove the normative requirement (and thus the word "must") completely? I'll brainstorm how that could/should read.

PR #411, opened 2015-07-01
RFC: Replace resource with response

devd:
Not sure if this is good or just makes it more confusing.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -237,7 +237,7 @@ result of the following command line:
 
 Conformant user agents MUST support the [SHA-256][sha2], [SHA-384][sha2]
 and [SHA-512][sha2] cryptographic hash functions for use as part of a
-response's [integrity metadata][], and MAY support additional hash functions.
+request's [integrity metadata][], and MAY support additional hash functions.
 
 <section>
 #### Agility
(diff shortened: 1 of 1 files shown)


devd:
@mnot can you also take a look?

@metromoxie @mozfreddyb @fmarier

fmarier:
I have to say, it does feel a little strange that the spec is called
"sub-resource integrity" and that we're not using the word "resource"
anymore...

devd:
heh .. I was sad about that too. fwiw, it is not replaced everywhere: this only changes it where I thought response made more sense than resource. there are still quite a few places where resource is used. The way I see it: the integrity is specified for the resource; it is checked on the response.

annevk:
Note that Fetch does have the concept of a "resource request", which I think is what you call "subresource", except it's grounded in first principles.

mnot:
LGTM (tho I'm on the road and only had a quick look). Thanks.

specs/subresourceintegrity/spec.markdown
```diff
@@ -659,10 +659,10 @@ Replace step 14.1 of HTML5's ["prepare a script" algorithm][prepare] with:
 ## Proxies
 
 Optimizing proxies and other intermediate servers which modify the
-content of fetched resources MUST ensure that the digest associated
-with those resources stays in sync with the new content. One option
+response MUST ensure that the digest associated
```
metromoxie: The pluralization doesn't match up here. Probably should read "...which modify the responses MUST ensure that..." (note added "s" to "response").

metromoxie:
This generally lgtm (with one nit I commented on). However, I would say that I feel OK with using the term "resource" in the non-normative sections since that is the colloquial term often used. Certainly safer to go with "response" though. We might even add a note somewhere explaining the term "Subresource Integrity" and the use of "response" :-)

devd:
thanks! I made the pluralization. I think we can add the note later if we want.

PR #420, opened 2015-07-10
Fixing some markup errors in the charter

jonathanKingston:
The `Web Applications Working Group/dt>` markup was visible for me in chrome... conditionally (which was odd).

However anyway fixing some broken bits of HTML however still wouldn't validate etc.


mozfreddyb:
looks good to me.
@hillbrad can we just merge this?

jonathanKingston:
@hillbrad thanks :+1:

Issue #418, opened 2015-07-09
SRI: explain CORS requirement clearly


jonathanKingston:
The CORS requirement in the specification I feel isn't clear enough for developers to understand why CORS was made as a requirement.

This comment clearly explains the reason it was added:
https://github.com/w3c/webappsec/issues/338#issuecomment-99766294

A clear example of an attacker being able to bypass firewall security by:
- loading local URLs until they stop triggering hash errors
- using the hash to calculate the content of the file

Clearly explaining that:
  This attack bypasses the current security of the attacker only able to execute the code instead of being able to know all of its contents.

metromoxie:
cc @devd @fmarier @mozfreddyb, and maybe of interest to @annevk and @mikewest

jonathanKingston:
This came about from a discussion raised on trying to get Ember to default SRI to be enabled by default which is risky if it is deployed over multiple origins.

The only safe thing I can get my [addon](https://www.npmjs.com/package/ember-cli-sri) to do is to only add integrities to relative URLs unless the user explicitly sets a crossorigin. (As mentioned in previous bugs this will harm implementations although a solution isn't clear either).

The only next best thing is to explain clearly to developers and library writers why CORS is required for assets on their servers.

annevk:
I'm not sure this needs additional clarification. It's no different from cross-origin scripts not throwing useful exceptions unless CORS is used.

annevk:
But also, SRI is more generic than just scripts.

jonathanKingston:
@annevk I would argue it could be overlooked by a new programmer being forced to implement SRI. Clearly explaining why things are harder than they might expect should lower the barrier to them giving up etc.

Is there was a more generic place like the CORS spec itself this one could cite that might be better for future specs also?

annevk:
Hmm, https://fetch.spec.whatwg.org/#http-cors-protocol just has a basic explanation of what it protects. If anything `<script>` should elaborate on how it hacks around CORS by executing opaque responses.

fmarier:
IMHO, we should re-phrase section 5.3 under Security Considerations to cover what Jonathan is asking for. I've had to explain it to a few people already, so I don't think it will be non-obvious to a lot of readers of this spec.

Additionally, the last sentence of that section no longer makes sense:

> User agents SHOULD mitigate the risk by refusing to fire error events on elements which loaded non-CORS cross-origin resources, but some side-channels will likely be difficult to avoid.

[...]

PR #422, opened 2015-07-12
Remove note about non CORS load

devd:


fmarier:
It seems to me like that whole section is basically trying to explain what #418 was asking about but doing a poor job at it.

We should either remove it entirely or make it useful to people who need a few more pointers about what CORS does and what attacks it prevents. My preference would be for the latter since I've had lots of questions about CORS when presenting SRI to developers.

fmarier:
And to be clear, we're doing a poor job at this because this section predates the CORS requirement, so it was probably fine before then :)

mozfreddyb:
Let's merge this and discuss about additional pointers in #418.

devd:
my thinking exactly: I think this is obviously wrong so we can remove it and then make additional decisions about this section.

PR #426, opened 2015-07-14
SRI: Changing rel value to keyword for matching stylesheets

jonathanKingston:
As rel can contain multiple values using the word keyword would allow for different attributes here.


mozfreddyb:
This is something that needs to be addressed in implementations, if they match `relValue == "stylesheet"`.
@fmarier said he's already doing the right thing. @metromoxie, can you take a look?

fmarier:
lgtm

fmarier:
We should also find a way to test it in the web platform tests. /cc @hillbrad

metromoxie:
I'll check tomorrow.

metromoxie:
It appears that Chrome is behaving correctly here. I'll update the web platform tests shortly with some tests for this.

jonathanKingston:
:+1: thanks @metromoxie, @fmarier and @mozfreddyb for looking into this.

PR #431, opened 2015-07-19
SRI: Francois' review of the spec

fmarier:
Here are a bunch of small fixes and editorial changes to the spec I'd like to propose after reading through the whole spec.

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -29,9 +29,9 @@ impunity. Ideally, authors would not only be able to pin the keys of a
 server, but also pin the _content_, ensuring that an exact representation of
 a resource, and _only_ that representation, loads and executes.
 
-This document specifies such a validation scheme, extending two HTML
-elements with an `integrity` attribute that contains a cryptographic hash of
-the representation of the resource the author expects to load. For instance,
+This document specifies such a validation scheme, extending two HTML elements
+and the `fetch()` API with an `integrity` attribute that contains a cryptographic hash
+of the representation of the resource the author expects to load. For instance,
 an author may wish to load some framework from a shared server rather than hosting it
 on their own origin. Specifying that the _expected_ SHA-256 hash of
 `https://example.com/example-framework.js`
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -675,7 +675,7 @@ with a value of [`no-transform`][notransform].
 <section>
 ### Non-secure contexts remain non-secure
 
-[Integrity metadata][] delivered to a context that is not a [secure context],
+[Integrity metadata][] delivered by a context that is not a [secure context],
 such as an HTTP page, only protects an origin against a compromise of the
 server where an external resources is hosted. Network attackers can alter the
 digest in-flight (or remove it entirely, or do absolutely anything else to the
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -60,7 +60,7 @@ and future versions of the specification are likely to expand this coverage.
 <section>
 ### Goals
 
-1.  Compromise of the third-party service should not automatically mean
+1.  Compromise of a third-party service should not automatically mean
     compromise of every site which includes its scripts. Content authors
     will have a mechanism by which they can specify expectations for
     content they load, meaning for example that they could load a
@@ -277,7 +277,7 @@ stronger hash functions as they become available.
 <section>
 #### Priority
 
-User agents must provide a mechanism of determining the relative priority of two
+User agents must provide a mechanism for determining the relative priority of two
 hash functions and return the empty string if the priority is equal. That is, if
 a user agent implemented a function like <dfn>getPrioritizedHashFunction(a,
 b)</dfn> it would return the hash function the user agent considers the most
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -328,12 +328,12 @@ resources accessed over a `file` scheme URL are unlikely to be
 eligible for integrity checks.
 {:.note}
 
-One should note that being a [secure document][] (e.g. a document delivered
+One should note that being a [secure document][] (e.g., a document delivered
 over HTTPS) is not necessary for the use of integrity validation. Because
 resource integrity is only an application level security tool, and it does not
 change the security state of the user agent, a [secure document] is
 unnecessary. However, if integrity is used in something other than a [secure document][]
-(e.g. a document delivered over HTTP), authors should be aware that the integrity
+(e.g., a document delivered over HTTP), authors should be aware that the integrity
 provides <em>no security guarantees at all</em>. For this reason, authors should
 only deliver integrity metadata on a [potentially secure origin][].  See
 [Non-secure contexts remain non-secure][] for more discussion.
@@ -598,7 +598,7 @@ integrity check <em>and</em> MUST return a network error, as described in
 [Modifications to Fetch][].
 
 On a failed integrity check, an <code>error</code> event is thrown. Developers
-wishing to provide a canonical fallback resource (e.g. a resource not served
+wishing to provide a canonical fallback resource (e.g., a resource not served
 from a CDN, perhaps from a secondary, trusted, but slower source) can catch this
 <code>error</code> event and provide an appropriate handler to replace the
 failed resource with a different one.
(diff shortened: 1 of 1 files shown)


specs/subresourceintegrity/spec.markdown
```diff
@@ -576,7 +576,7 @@ as possible.
 <section>
 ##### HTMLLinkElement
 
-attribute DOMString integrity
+readonly attribute DOMString integrity
```
annevk: Why would you not be able to set it?

specs/subresourceintegrity/spec.markdown
```diff
@@ -739,8 +739,9 @@ Much of the content here is inspired heavily by Gervase
 Markham's [Link Fingerprints][] concept, as well as WHATWG's [Link Hashes][].
 
 A special thanks to Mike West of Google, Inc. for his invaluable contributions
-to the initial version of this spec. Additonally, Brad Hill, Anne van Kesteren, Mark Nottingham, 
-Dan Veditz, Eduardo Vela, Tanvi Vyas, and Michal Zalewski provided invaluable feedback.
+to the initial version of this spec. Additonally, Brad Hill, Jonathan Kingston,
+Anne van Kesteren, Mark Nottingham, Dan Veditz, Eduardo Vela, Tanvi Vyas, and
```
annevk: "Ke" sorts before "Ki".

specs/subresourceintegrity/spec.markdown
```diff
@@ -332,8 +332,8 @@ One should note that being a [secure document][] (e.g. a document delivered
 over HTTPS) is not necessary for the use of integrity validation. Because
 resource integrity is only an application level security tool, and it does not
 change the security state of the user agent, a [secure document] is
-unnecessary. However, if integrity is used in other than a [secure document][] (e.g.
-a document delivered over HTTP), authors should be aware that the integrity
+unnecessary. However, if integrity is used in somethig else than a [secure document][]
```
annevk: Typo "somethig"

specs/subresourceintegrity/spec.markdown
```diff
@@ -332,8 +332,8 @@ One should note that being a [secure document][] (e.g. a document delivered
...
+(e.g. a document delivered over HTTP), authors should be aware that the integrity
```
annevk: Requires a comma after e.g. I think.

specs/subresourceintegrity/spec.markdown
```diff
@@ -29,7 +29,7 @@ impunity. Ideally, authors would not only be able to pin the keys of a
 server, but also pin the _content_, ensuring that an exact representation of
 a resource, and _only_ that representation, loads and executes.
 
-This document specifies such a validation scheme, extending several HTML
+This document specifies such a validation scheme, extending two HTML
```
annevk: And apparently extending the `fetch()` API?

specs/subresourceintegrity/spec.markdown
jonathanKingston: alternate styles would be attackable by scripts perhaps?

specs/subresourceintegrity/spec.markdown
```diff
@@ -29,9 +29,9 @@ impunity. Ideally, authors would not only be able to pin the keys of a
 server, but also pin the _content_, ensuring that an exact representation of
 a resource, and _only_ that representation, loads and executes.
 
-This document specifies such a validation scheme, extending several HTML
-elements with an `integrity` attribute that contains a cryptographic hash of
-the representation of the resource the author expects to load. For instance,
+This document specifies such a validation scheme, extending two HTML elements
+and the fetch() API with an `integrity` attribute that contains a cryptographic hash
```
fmarier: @annevk is that the kind of wording you had in mind?

specs/subresourceintegrity/spec.markdown
fmarier: Are you sure? Everywhere in the spec we use "e.g." without a comma after it.

specs/subresourceintegrity/spec.markdown
fmarier: We have it as `readonly` in Section 3.4 (modifications to Fetch) too.

I thought that was to indicate that it can only be read and never modified. I can't think of a reason why we'd want to modify it. Are you thinking perhaps of add-ons being able to change it?

jonathanKingston: @fmarier how would you protect it from being abused if it wasn't readonly?

specs/subresourceintegrity/spec.markdown
mozfreddyb: I learned to put a comma after both, but I'm not a native speaker \*shrugs\*.
Reading through https://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e I'd still go for a comma everywhere :-)

fmarier: And is there a difference when it's inside parentheses though? This [Oatmeal comic](http://theoatmeal.com/comics/ie) (I know, not the most authoritative source) seems to say there is:

An example of e.g.

specs/subresourceintegrity/spec.markdown
annevk: You can invoke `setAttribute()` too so I don't think it makes much sense. For Fetch' `Request` readonly makes sense, since `Request`s are mostly immutable.

specs/subresourceintegrity/spec.markdown
annevk: Seems alright. I would probably use \` around `fetch()`.

specs/subresourceintegrity/spec.markdown
annevk: Those Oatmeal guidelines seem fairly confusing. @sideshowbarker told me to use a comma always.

sideshowbarker: > > Those Oatmeal guidelines seem fairly confusing. @sideshowbarker told me to use a comma always.

> Are you sure? Everywhere in the spec we use "e.g." without a comma after it.

Yeah, I'm sure on this one. Major style guides such as the MLA (Modern Language Association) and APA (American Psychological Association) style guides (the main style guides used by many academic publications) and the Chicago Manual of Style (used by most US newspapers and magazines and many other places) all say that "e.g." should always be followed by a comma.

The text of those guides and most other major style guides isn't available online but did find this:
 
http://blog.apastyle.org/apastyle/2010/05/its-all-latin-to-me.html

Also FWIW I worked professionally as a technical writer at two different software companies for a number of years and I can assert that in my experience the standard style in the technical-writing industry is also to always follow "e.g." with a comma.

specs/subresourceintegrity/spec.markdown
```diff
@@ -243,7 +243,7 @@ request's [integrity metadata][], and MAY support additional hash functions.
 #### Agility
 
 Multiple sets of [integrity metadata][] may be associated with a single
-resource in order to provide agility in the face of future discoveries.
+resource in order to provide agility in the face of future cryptographic developments.
```
devd: I am not sure about this change; breaking a hash function is not quite a cryptographic development as much as a discovery. but don't really care.

specs/subresourceintegrity/spec.markdown
```diff
@@ -328,12 +328,12 @@ resources accessed over a `file` scheme URL are unlikely to be
...
 {:.note}
 
-One should note that being a [secure document][] (e.g. a document delivered
+One should note that being a [secure document][] (e.g., a document delivered
 over HTTPS) is not necessary for the use of integrity validation. Because
 resource integrity is only an application level security tool, and it does not
 change the security state of the user agent, a [secure document] is
-unnecessary. However, if integrity is used in other than a [secure document][] (e.g.
-a document delivered over HTTP), authors should be aware that the integrity
+unnecessary. However, if integrity is used in something else than a [secure document][]
```
devd: I think this should be "something other than"

specs/subresourceintegrity/spec.markdown
```diff
@@ -676,7 +676,7 @@ with a value of [`no-transform`][notransform].
 ### Non-secure contexts remain non-secure
 
 [Integrity metadata][] delivered to a context that is not a [secure context],
```
devd: should this by "delivered by a context that is not a .."

specs/subresourceintegrity/spec.markdown
```diff
@@ -534,7 +534,7 @@ value each element's `integrity` content attribute is added to the
 
 A future revision of this specification is likely to include integrity support
 for all possible subresources, i.e., `a`, `audio`, `embed`, `iframe`, `img`,
-`link`, `object`, `script`, `source`, `track`, and `video` elements.
+`object`, `script`, `source`, `track`, and `video` elements.
```
devd: there are other types of link tags too and we only do styles currently. So not sure if we should remove it: see, e.g., https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

fmarier:
@devd I've addressed all of your comments. Ready to merge?

devd:
lgtm

Issue #409, opened 2015-06-29
Referrer attribute and Fetch integration


annevk:
As I said on blink-dev I'm not convinced this attribute with only a limited set of values is the way to go. We need to be able to set arbitrary (same-origin) URLs if we want referrers to remain a useful feature in the world of service workers.

E.g. if you have a worker at URL A and a service worker at URL B, the service worker needs to be able to override the referrer to be A for its outgoing requests as otherwise when a site migrates to service workers all its referrers will be B, which would be different from what it had before.

jeisinger:
I think we should remove it from the draft, @mikewest wdyt

mikewest:
Sure. No one's implemented it, the only partial implementation had negative feedback on blink-dev@, etc. I think it's fine to drop the feature from the current spec, and come up with something that suits the actual use-cases this is supposed to serve.

/cc @fmarier, who added it in the first place (in w3c/webappsec#175). It would be worth poking at https://lists.w3.org/Archives/Public/public-webappsec/2015Feb/0228.html since some folks there were enthusiastic (including me, but I'm enthusiastic about too many things these days...).

fmarier:
On the gecko side, @franziskuskiefer has already landed support for the attribute in [`<img>`](https://bugzilla.mozilla.org/show_bug.cgi?id=franziskuskiefer1166910) and has patches for [`<a>` and `<area>`](https://bugzilla.mozilla.org/show_bug.cgi?id=1174913) and [`<iframe>`](https://bugzilla.mozilla.org/show_bug.cgi?id=1175736) in review.

briansmith:
I don't think it's desirable for a ServiceWorker to set the referrer attribute on any request, given the advertising use cases for Referer that Mike West gave on the mailing list a few months ago. In particular, didn't Mike say that it was important that an advertisement be able to get the full referrer URL without any tampering? Otherwise, a ServiceWorker could be used to trick the ad network into showing an ad on a page that the ad network doesn't want to show an ad on. It seems like the path information, at least, in a Referer header isn't useful, in general, when a ServiceWorker is involved.

Anyway, the referrer attribute is essential for allowing pages and browsers to default to sending no referrer, while still allowing the page to send the referrer for specific advertisements that need it. We could (ab)use link relations for this purpose, but most elements that need such referrer control do not currently have a rel attribute.

jeisinger:
we already have rel=noreferrer. also, adding a referrer attribute doesn't cover programmatic navigations.

[...]

PR #435, opened 2015-07-23
rename 'referrer' attribute to 'referrerpolicy', make rel=noreferrer take priority

sc0ttbeardsley:
@annevk @jeisinger this is an attempt to address the comments regarding renaming "referrer" to "referrerpolicy" as discussed in https://github.com/w3c/webappsec/issues/409

I see there are other changes being added but I only did a "make referrer-policy/index.html" so maybe that is expected?

/cc @wooseok

### DIFF
--- a/specs/referrer-policy/index.html
+++ b/specs/referrer-policy/index.html
@@ -70,8 +70,8 @@
   
    <h1 class="p-name no-ref" id="title">Referrer Policy</h1>
   
-   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft
-    <time class="dt-updated" datetime="2015-05-11">11 May 2015</time></span></h2>
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Editor’s Draft,
+    <time class="dt-updated" datetime="2015-07-22">22 July 2015</time></span></h2>
   
    <div data-fill-with="spec-metadata">
     <dl>
@@ -82,7 +82,7 @@
      <dt>Version History:
      <dd><a href="https://github.com/w3c/webappsec/commits/master/specs/referrer-policy/index.src.html">https://github.com/w3c/webappsec/commits/master/specs/referrer-policy/index.src.html</a>
      <dt>Feedback:
-     <dd><span><a href="mailto:public-webappsec@w3.org?subject=%5BREFERRER%5D%20feedback">public-webappsec@w3.org</a> with subject line “<kbd>[REFERRER] <var>… message topic …</var></kbd>” (<a href="http://lists.w3.org/Archives/Public/public-webappsec/" rel="discussion">archives</a>)</span>
+     <dd><span><a href="mailto:public-webappsec@w3.org?subject=%5BREFERRER%5D%20YOUR%20TOPIC%20HERE">public-webappsec@w3.org</a> with subject line “<kbd>[REFERRER] <var>… message topic …</var></kbd>” (<a href="http://lists.w3.org/Archives/Public/public-webappsec/" rel="discussion">archives</a>)</span>
      <dt>Issue Tracking:
      <dd><a href="#issues-index">Inline In Spec</a>
      <dt class="editor">Editors:
@@ -184,7 +184,7 @@
        </ul>
       <li><a href="#referrer-policy-delivery-meta"><span class="secno">4.2</span> <span class="content">Delivery via <code><span>meta</span></code></span></a>
       <li><a href="#referrer-policy-delivery-referrer-attribute"><span class="secno">4.3</span> <span class="content">Delivery
-  via a <code>referrer</code> attribute</span></a>
+  via a <code>referrerpolicy</code> attribute</span></a>
       <li><a href="#referrer-policy-delivery-implicit"><span class="secno">4.4</span> <span class="content">Implicit Delivery</span></a>
        <ul class="toc">
         <li><a href="#referrer-policy-delivery-implicit-nested"><span class="secno">4.4.1</span> <span class="content">
@@ -318,9 +318,7 @@
       <p>If no referrer policy is explicitly set for a <a data-link-type="dfn" href="http://www.w3.org/TR/html5/webappapis.html#settings-object">settings object</a>,
       then the value of the property is <code>null</code>. Otherwise, the value
       is whatever has been explicitly set, as explained in the
-      <a href="#set-referrer-policy">§6.1 
-    Set environment’s referrer policy to policy
-  </a> algorithm.</p>
+      <a href="#set-referrer-policy">§6.1 Set environment’s referrer policy to policy</a> algorithm.</p>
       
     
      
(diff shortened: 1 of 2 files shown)


specs/referrer-policy/index.html
```diff
@@ -1419,21 +1405,21 @@ <h3 class="no-ref no-num heading settled" id="conformance-classes"><span class="
...
+   <li><a href="#no-referrer-when-downgrade">No Referrer When Downgrade</a><span>, in §3.2</span>
+   <li><a href="#origin-only">Origin Only</a><span>, in §3.3</span>
+   <li><a href="#origin_only-flag">origin-only flag</a><span>, in §6.3</span>
+   <li><a href="#origin-when-cross_origin">Origin When Cross-Origin</a><span>, in §3.4</span>
+   <li><a href="#referrer-policy">policy</a><span>, in §2</span>
+   <li><a href="#referrer-directive">referrer directive</a><span>, in §4.1</span>
+   <li><a href="#referrer-policy">referrer policy</a><span>, in §2</span>
+   <li><a href="#same_origin-request">same-origin</a><span>, in §2</span>
+   <li><a href="#same_origin-request">same-origin request</a><span>, in §2</span>
+   <li><a href="#unsafe-url">Unsafe URL</a><span>, in §3.5</span></ul>
```
sc0ttbeardsley: for instance this.. maybe bikeshed pulled this in?

mikewest:
I'll defer to @jeisinger here as he was involved in the conversation about this attribute more than I was, but this looks pretty reasonable to me.

Also, you're correct that most of the `index.html` changes are indeed just new Bikeshed updates. Bikeshed moves quickly. ;)

specs/referrer-policy/index.src.html
```diff
@@ -516,18 +516,18 @@ <h3 id="referrer-policy-delivery-meta">Delivery via <{meta}></h3>
...
   element, for example:
 
   <pre class="example">
-    &lt;a href="http://example.com" referrer="origin"&gt;
+    &lt;a href="http://example.com" referrerpolicy="origin"&gt;
   </pre>
 
-  The following values for the <code>referrer</code> attribute are valid,
+  The following values for the <code>referrerpolicy</code> attribute are valid,
   and map to the listed <a>referrer policy</a> values:
```
jeisinger: i'd rather have all referrer policies that are accepted in the meta tag to be valid values, so we can use it for feature detection.

Also, the attribute should be spec'd to to reflect its value and be restricted to valid values.

specs/referrer-policy/index.src.html
```diff
@@ -539,15 +539,15 @@ <h3 id="referrer-policy-delivery-referrer-attribute">Delivery
...
   takes precedence over the policy defined for the whole document via CSP or
   a <a element>meta</a> element.
 
   ISSUE: If an <code><a element>a</a></code>
   or <code><a element>area</a></code> element includes both
-  a <code>referrer</code> attribute as well as
+  a <code>referrerpolicy</code> attribute as well as
   a <a href="https://html.spec.whatwg.org/#link-type-noreferrer"><code>noreferrer</code>
-  link type</a>, should the policy of the <code>referrer</code> attribute
+  link type</a>, should the policy of the <code>referrerpolicy</code> attribute
```
jeisinger: i think we should honor the rel=noreferrer btw

specs/referrer-policy/index.src.html
```diff
@@ -516,18 +516,18 @@ <h3 id="referrer-policy-delivery-meta">Delivery via <{meta}></h3>
   </dl>
 
   <h3 id="referrer-policy-delivery-referrer-attribute">Delivery
-  via a <code>referrer</code> attribute</h3>
+  via a <code>referrerpolicy</code> attribute</h3>
 
-  A referrer policy may be set when a <code>referrer</code> attribute is
+  A referrer policy may be set when a <code>referrerpolicy</code> attribute is
```
jeisinger: the spec should also include what happens. e.g. if the user clicks on the link, but it has an onclick handler that then creates an XHR, does the referrer policy apply to the XHR as well?

What if it uses window.open() to navigate, or assign a value to location.href?

annevk: I don't think either of those should really be affected by this attribute on an `<a>`. That would be way too magical.

specs/referrer-policy/index.src.html
sc0ttbeardsley: done

specs/referrer-policy/index.src.html
sc0ttbeardsley: I like magic, but ya, might be a bit out of scope.

specs/referrer-policy/index.src.html
sc0ttbeardsley: @jeisinger done, taking the conservative approach: https://github.com/sc0ttbeardsley/webappsec/commit/67018e17b0cf52a3decb609a2105282f7a7817ae

sc0ttbeardsley:
anything else to do here or can it be merged?

specs/referrer-policy/index.src.html
jeisinger: can you also add that the attribute should be reflecting only valid policies.

And what should document.createElement('a').referrerpolicy be? Nothing? The current referrer policy?

specs/referrer-policy/index.src.html
```diff
@@ -535,24 +537,24 @@ <h3 id="referrer-policy-delivery-referrer-attribute">Delivery
...
-  link type</a>, should the policy of the <code>referrer</code> attribute
-  take precedence over the link type, as suggested
-  by <a href="https://lists.w3.org/Archives/Public/public-webappsec/2015Feb/0235.html">Martin
-  Thomson</a>, or should we take the more conservative approach as suggested
-  by <a href="https://lists.w3.org/Archives/Public/public-webappsec/2015Feb/0283.html">Brian
-  Smith</a> and honor the <code>noreferrer</code> link type?
+  a <code>referrerpolicy</code> attribute as well as
+  a <a><code>noreferrer</code></a> link type then the <a><code>noreferrer</code></a>
+  link type will take precedence and the <a><code>No Referrer</code></a> policy
+  takes effect.
```
jeisinger: can you add some text to what the referrer policy applies? Something like the request?

I guess for <a> and <area> it's the navigation triggered by the user agent when clicking on it.

for  it's the request for src

what about <iframe>, is it for loading the iframe only, or will it also set the policy for the iframe's global context?

specs/referrer-policy/index.src.html
annevk: In https://fetch.spec.whatwg.org/#referrerpolicy I defined a default "" value. I think we should name the IDL attribute `referrerPolicy` btw. It's normal for IDL attributes to have proper casing.

specs/referrer-policy/index.src.html
annevk: For `<iframe>` seems weird because e.g., `<a>` can also initiate a load for an `<iframe>`. How would that go together?

franziskuskiefer:
@sc0ttbeardsley any progress on this?

sc0ttbeardsley:
@franziskuskiefer i will try to address some of the comments but many are unrelated to the original intent of the PR (to rename 'referrer' attribute to 'referrerpolicy')

sc0ttbeardsley:
I've moved this request to the new repo (and have limited it to only the referrer -> referrerpolicy changes to avoid some scope creep). https://github.com/w3c/webappsec-referrer-policy/pull/1

PR #437, opened 2015-07-25
Fail closed on CORS failures

metromoxie:
This is in response to the apparent consensus in #317. This fails closed on CORS failures so that it is clear to developers that integrity cannot be used without CORS.


metromoxie:
cc @mozfreddyb @fmarier @devd to review.

devd:
To summarize, all this does is that if you specify integrity attribute on a
crossorigin load then cors error leads to block?

What about a cross origin resource that redirects to a same origin resource
and wasn't a cors enabled fetch?

jonathanKingston:
:+1: looks good to me.

A redirect should resolve before the integrity checking anyway right?

metromoxie:
I think @jonathanKingston is correct, but @annevk can you verify? That was certainly one of the intents of the Fetch modifications, at any rate.

fmarier:
Based on these modifications, all of these would be loaded:

```
<script src="http://elsewhere.example.com/foo.js" integrity="">
<script src="http://elsewhere.example.com/foo.js" integrity="        ">
<script src="http://elsewhere.example.com/foo.js" integrity="rot13-base64hash">
<script src="http://elsewhere.example.com/foo.js" integrity="invalid">
```

because they all correspond to `no-metadata` and will get the early `true` return.

I think the last two should be blocked. For this, we could add an extra variable in the parse metadata algorithm to keep track of whether the `integrity` attribute is empty (i.e. consists only of whitespace) or whether is it non-empty. Maybe something like:

1. Let `result` be the empty set.
2. **Let `empty` be equal to `true`.**
3. For each `token` returned by splitting `metadata` on spaces:
  0. **Set `empty` to `false`**
  1. If `token` is not a valid metadata, skip the remaining steps, and proceed to the next token.
  2. Parse `token` per the grammar in `integrity metadata`
  3.Let `algorithm` be the alg component of `token`.
  4. If `algorithm` is a hash function recognized by the user agent, add the parsed token to `result`.
3. **Return `no metadata` if `empty` is true, otherwise return `result`.**

with a step that looks for empty set of metadata in the main algorithm:

1. Let `parsedMetadata` be the result of parsing `metadataList`.
2. If `parsedMetadata` is `no metadata`, return `true`.
3. If response is not eligible for integrity validation, return `false`.
4. **If `parsedMetadata` is the empty set, return `true`.**
5. Let `metadata` be the result of getting the strongest metadata from `parsedMetadata`.

metromoxie:
I disagree, as I don't see why we'd special case a single value as a parse error, when if it was present in a metadata list it would not cause a parse error (for example, if you had `integrity="sha256-deadbeef rot13-somehashvalue"`). To me, that seems even more odd. Also, this prevents syntax changes in the future where this is valid syntax.

fmarier:
> I disagree, as I don't see why we'd special case a single value as a parse error, when if it was present in a metadata list it would not cause a parse error (for example, if you had integrity="sha256-deadbeef rot13-somehashvalue")

Just to be clear, I'm suggesting we treat `integrity="invalidgarbage"` and `integrity="rot13-hashvalue"` the same way:
- ignored by the parser
- and not equal to `integrity=""` or `integrity="   "`

So if you have `<script ... integrity="sha256-hash rot13-hash" cross-origin="anonymous">` then we ignore `rot13-hash` and do the load (assuming the `sha256-hash` checks out). But if you have `<script ... integrity="sha256-hash rot13-hash">` then we block it because of missing CORS. Take out the `sha256-hash` from the above and the behavior doesn't change.

specs/subresourceintegrity/spec.markdown
```diff
@@ -412,7 +412,7 @@ the user agent.
     [parsing <var>metadataList</var>][parse].
 2.  If <var>parsedMetadata</var> is `no metadata`, return `true`.
 3.  If [<var>response</var> is not eligible for integrity
```
annevk: What does "not eligible" mean? It doesn't seem to be a reference.

annevk:
Per Fetch integrity checking happens on the final response (after redirects have been followed). It's not clear to me though that this change makes sense, as it effectively changes the default to "same-origin", but seems to do so in a rather poor way.

specs/subresourceintegrity/spec.markdown
fmarier: It refers to [this algorithm](https://w3c.github.io/webappsec/specs/subresourceintegrity/#is-varresponsevar-eligible-for-integrity-validation) which says it needs to be CORS or same-origin to be eligible.

metromoxie: I added in a link to the 'is eligible' algorithm to clarify.

metromoxie:
@fmarier: But if you have `<script ... integrity="sha256-hash rot13-hash" crossorigin="anonymous">` and `<script ... integrity="rot13-hash" crossorigin="anonymous">`, the first will not throw an error and the second will. That seems like an arbitrary difference to me, and that's why I don't like it. In any case, if you would like to continue a discussion on this, may I suggest a new issue, as I think it's a separate decision from this.

fmarier:
> But if you have \<script ... integrity="sha256-hash rot13-hash" crossorigin="anonymous"\> and \<script ... integrity="rot13-hash" crossorigin="anonymous"\>, the first will not throw an error and the second will.

Both of these will load since they are both eligible.

The first one will load because the sha256 hash checks out, the second one will load because of the extra step I added to the main algorithm:

`4. If parsedMetadata is the empty set, return true.`

metromoxie:
@fmarier Ah, yes, my mistake. However, this really seems like an unnecessary complication of the algorithm to me. IMO, unparsable/unrecognized algorithms should just be treated as empty. In any case, like I said, I'd like to treat this as a separate issue.

fmarier:
> @fmarier Ah, yes, my mistake. However, this really seems like an unnecessary complication of the algorithm to me. IMO, unparsable/unrecognized algorithms should just be treated as empty. In any case, like I said, I'd like to treat this as a separate issue.

Ok, I'll submit as a separate PR once this one is merged. I do think it's important to fail closed when we don't have to worry about backward/forward compatibility.

We haven't had much feedback on the list about this PR, time to merge?

devd:
sgtm

PR #446, opened 2015-08-05
Fix typo

taravancil:


mikewest:
Thanks for the PR! Unfortunately, this is a historical artifact! We published it with the typo, so, we're stuck with it forever. :/

Happily, @jonathanKingston fixed the typo in the source document, so we won't have the typo in the future.

PR #452, opened 2015-08-12
Credential Manager API. Use store instead of notifySignIn.

tehhi:


mikewest:
Sorry for the churn. :/

PR #455, opened 2015-08-14
SRI: re-introduce the distinction between empty and invalid metadata

fmarier:
This commit reintroduces a distinction that used to exist in the spec prior to 930ec20d816f20e6279a6e972e101e457101d948 (see #86 and #119).

As [discussed on the list](https://lists.w3.org/Archives/Public/public-webappsec/2015Aug/0006.html), it helps developers catch mistakes by failing closed on CORS errors when the `integrity` attribute is non-empty (an indication that the developer meant to use SRI).

Prior to this commit, `integrity` attributes which consists of only invalid metadata would fail open (silently load).

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -372,16 +372,18 @@ valid hash expressions whose hash functions are understood by
 the user agent.
 
 1.  Let <var>result</var> be the empty set.
-2.  For each <var>token</var> returned by [splitting <var>metadata</var> on
+2.  Let <var>empty</var> be equal to `true`.
+3.  For each <var>token</var> returned by [splitting <var>metadata</var> on
     spaces][split-on-spaces]:
-    1.  If <var>token</var> is not a valid metadata, skip the remaining
+    1.  Set <var>empty</var> to `false`.
+    2.  If <var>token</var> is not a valid metadata, skip the remaining
         steps, and proceed to the next token.
-    2.  Parse <var>token</var> per the grammar in [integrity metadata][]
-    3.  Let <var>algorithm</var> be the <var>alg</var> component of
+    3.  Parse <var>token</var> per the grammar in [integrity metadata][]
+    4.  Let <var>algorithm</var> be the <var>alg</var> component of
         <var>token</var>.
-    4.  If <var>algorithm</var> is a hash function recognized by the user
+    5.  If <var>algorithm</var> is a hash function recognized by the user
         agent, add the parsed <var>token</var> to <var>result</var>.
-3.  Return `no metadata` if <var>result</var> is empty, otherwise return
+4.  Return `no metadata` if <var>empty</var> is `true`, otherwise return
     <var>result</var>.
 
 [split-on-spaces]: http://www.w3.org/TR/html5/infrastructure.html#split-a-string-on-spaces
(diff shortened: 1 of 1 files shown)


fmarier:
Note: @metromoxie [asked](https://github.com/w3c/webappsec/pull/437#issuecomment-125678364) for this to be done as a separate issue. It addresses the remaining questions from #317.

devd:
hmm.. can you explain more on how this achieves that? At a glance, the changes seem very orthogonal and some handholding on what changes you made and why will help make sure that everyone is in sync.  It is the spec version of adding lots of comments in the code ;)

fmarier:
The main idea behind this change is that *parsing the metadata* will go from having two states:

1. **empty or invalid**
2. **valid**

to having three states:

1. **empty**: when the `integrity` attribute is empty or only contains whitespace (e.g. `integrity="     "`)
2. **invalid**: non-empty but no valid metadata (e.g. `integrity="garbage"`)
3. **valid**: there is at least one valid metadata found (e.g. `integrity="sha256-lkbjlkjklbv..."`)

This allows us to fail:

- **open** on empty metadata
- **closed** where there's no valid metadata (but the integrity attribute is not empty)

when doing a cross-origin load without the `crossorigin` attribute.

If the `crossorigin` attribute is present and there is only invalid metadata, we still fail open (for forward-compatibility) and this PR is not changing this.

fmarier:
I added a new `empty` variable in the "parse metadata" algorithm to go from two states to three states during the parsing of the attribute.

Because the early exit in the main algorithm only applies to the `no metadata` state, which is now restricted to empty metadata, I had to add a second early exit after the CORS check. That second early exit makes the "only invalid metadata" state fail open when the CORS attribute is correctly present.

devd:
I like the separating out "empty" and "invalid" into two distinct categories.

still confused. .. why should there be a cross origin attribute for same origin loads?

fmarier:
> still confused. .. why should there be a cross origin attribute for same origin loads?

There's no need for one, same-origin loads are automatically *eligible for SRI* without the need for the *crossorigin* attribute.

devd:
ok .. lgtm @metromoxie can you take a look too?

mozfreddyb:
sounds good to me too. WDYT, @metromoxie?

metromoxie:
As I've stated before, I don't like this change, as it seems somewhat arbitrary to allow integrity="somethingvalid somethinginvalid" but to reject integrity="somethinginvalid". But it seems I'm outvoted.

fmarier:
> As I've stated before, I don't like this change, as it seems somewhat arbitrary to allow integrity="somethingvalid somethinginvalid" but to reject integrity="somethinginvalid".

Actually, this patch is about treating these two in the same way. The idea is to reject both `integrity="somethingvalid somethinginvalid"` and `integrity="somethinginvalid"` **if the CORS attribute is missing** (and the load is cross-origin).

So we reject these:

    <script src="https://elsewhere.example.com/foo.js" integrity="sha256-valid garbage">
    <script src="https://elsewhere.example.com/foo.js" integrity="garbage">

and we allow this one:

    <script src="https://elsewhere.example.com/foo.js" integrity="">

devd:
per fmarier's comment, I think this is fine even for concerns that joel raised

PR #462, opened 2015-08-28
address review comments by timeless

hillbrad:
https://lists.w3.org/Archives/Public/public-webappsec/2015Aug/0107.html

### DIFF
--- a/specs/CSP2/index.html
+++ b/specs/CSP2/index.html
@@ -1,17 +1,9 @@
-<!doctype html>
-<html lang="en">
+<!doctype html><html lang="en">
  <head>
-  
   <meta content="text/html; charset=utf-8" http-equiv="Content-Type">
-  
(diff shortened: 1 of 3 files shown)


mikewest:
Thanks, Brad! The `index.src.html` changes LGTM, but the generated HTML changed more than I would have expected. I think we might have different versions of Bikeshed locally. If you're using tip-of-tree, then let's run with the changes. If you're using an older version, I'd suggest backing out the generated HTML changes, and I'll take care of them along with timeless' other concerns next week when I'm back in the office.

Thanks again!

hillbrad:
I do believe I'm on latest bikeshed as of yesterday.

mikewest:
Alright, then. LGTM to rebase and land. (Sorry I gave you conflicts. :( )

PR #466, opened 2015-09-05
COWL: intial commit towards FPWD

deian:
Thanks to @jonathanKingston for fixing a bunch of bugs in this!

### DIFF
--- a/specs/Makefile
+++ b/specs/Makefile
@@ -2,6 +2,7 @@ all: clean content-security-policy/index.html mixedcontent/index.html referrer-p
 
 clean:
 	rm -rf ./content-security-policy/index.html
+	rm -rf ./cowl/index.html
 	rm -rf ./credentialmanagement/index.html
 	rm -rf ./mixedcontent/index.html
 	rm -rf ./referrer-policy/index.html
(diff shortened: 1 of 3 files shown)


jonathanKingston:
@deian I can see a fair bit of difference with my bikeshed output have you updated that?

Also the generated output still has an instance of `labled` but the src file doesn't have that.

deian:
@jonathanKingston you're right, thanks for catching that -- didn't `make COWL`

mikewest:
Now's probably a good time to ask whether or not we want separate repositories for WebAppSec specs, or whether we'd like to keep them all in one repository (https://twitter.com/mikewest/status/639400263007236096).

/cc @hillbrad

mikewest:
(Happy to land this either way... do you have push rights to the repository?)

deian:
@mikewest sorry for delay. (On a boat, with periodic wiif-off periods.) I don't have push rights to the repo. Happy to keep it separate if we're going with that route. I can't tell from twritter convo if a decision was made.

mikewest:
Started a thread at https://lists.w3.org/Archives/Public/public-webappsec/2015Sep/0039.html (and CC'd you).

mikewest:
@wseltzer: I think the concensus is moving towards splitting the repos. Would you mind creating a COWL repository for Deian, and giving him push access, along with whoever else he'd like to help him edit the spec? I guess we could just reuse the webappsec group if you add him to it?

PR #470, opened 2015-09-11
Typo: "requirments" -> "requirements"

konklone:


konklone:
I only edited the `.src.html` file - `bikeshed` et al are very new to me. Happy to learn stuff if that's important.

mikewest:
This is totally good enough, thanks!

Issue #477, opened 2015-09-18
SRI: upgrade examples to sha384?


fmarier:
The NSA no longer recommends SHA-256 apparently: https://www.nsa.gov/ia/programs/suiteb_cryptography/index.shtml

Should we upgrade our examples to use `sha384` instead of `sha256`?

jonathanKingston:
Sounds good to me.
Time to support more algos too?

fmarier:
> Time to support more algos too?

I'd say no because that involves implementation changes too. Let's ship what we have now.

mozfreddyb:
> Should we upgrade our examples to use sha384 instead of sha256?

Yes!

jonathanKingston:
Sorry completely forgot level 1 hasn't shipped yet. (I meant more algos for
level 2 just to clarify)

fmarier:
> Sorry completely forgot level 1 hasn't shipped yet. (I meant more algos for level 2 just to clarify)

Yeah, that would be an easy V2 feature.

mozfreddyb:
Taking the sha256 → sha384 rewrite

mozfreddyb:
My patch is nearly done, mostly search&replace work, but I have a few questions..

I guess it's too late for SRIv1 to kill SHA256 completely, but we are arguing that SHA256 is a good thing in the document, when the NSA claims it is not anymore (and they probably know better?).

So I suppose we will have to keep this one?
> Conformant user agents MUST support the [SHA-256][sha2], [SHA-384][sha2]
> and [SHA-512][sha2] cryptographic hash functions for use as part of a
> request's [integrity metadata][], and MAY support additional hash functions.

And this one?
> Digests are only as strong as the hash function used to generate them. User
> agents SHOULD refuse to support known-weak hashing functions like MD5 or SHA-1,
> and SHOULD restrict supported hashing functions to those known to be
> collision-resistant. At the time of writing, SHA-256 is a good baseline.
> Moreover, user agents SHOULD re-evaluate their supported hash functions
> on a regular basis, and deprecate support for those functions shown to be
> insecure.

I'd rather not say that SHA256 really is a good baseline? Not sure whether we can still modify this sentence. Most tooling we have and CDNs we talk to unfortunately default to SHA256 by now :-/

fmarier:
> I'd rather not say that SHA256 really is a good baseline? Not sure whether we can still modify this sentence. Most tooling we have and CDNs we talk to unfortunately default to SHA256 by now :-/

[...]

PR #488, opened 2015-09-28
Rewrite instance of SHA256 to SHA384, see issue #477

mozfreddyb:
We discussed this in #477, so this should be merge-ready.


fmarier:
Looks good to me.

PR #491, opened 2015-09-29
Update date and status of Web Workers

AFBarstow:
This PR changes the Web Workers reference from the 2012 CR to the WD published on 24 September 2015.


hillbrad:
Hmm... did I misread the conclusion of the WebApps thread to reference the
WHATWG version in UPGRADE instead?  I'm happy that this is updated now, but
will it be maintained on an ongoing basis?

Mike?

mikewest:
This doesn't update the references, it just updates the doc we're referencing anyway, which makes total sense to me (now). :) Thanks!

PR #492, opened 2015-09-30
SRI: Add response tainting to the eligibility algorithm (fixes #487)

### LINKED ISSUE #487: SRI: add a note about data: and blob: URLs
To follow up on the issues raised in this thread: https://lists.w3.org/Archives/Public/public-webappsec/2015Sep/0172.html

I'm talking to @annevk and trying to come up with concrete test cases for this.

fmarier:
This was also raised on the mailing list:
https://lists.w3.org/Archives/Public/public-webappsec/2015Sep/0172.html

### DIFF
--- a/specs/subresourceintegrity/spec.markdown
+++ b/specs/subresourceintegrity/spec.markdown
@@ -350,19 +350,27 @@ The following algorithm details these restrictions:
 
 1.  Let <var>request</var> be the request that fetched
     <var>resource</var>.
-2.  If the [mode][fetch-mode] of <var>request</var> is `CORS`,
+2.  If the [response type][] is `opaque`, return `false`.
+3.  If the [mode][fetch-mode] of <var>request</var> is `CORS`,
     return `true`.
-3.  If the [origin][fetch-origin] of <var>request</var> is
+4.  If the [origin][fetch-origin] of <var>request</var> is
     <var>resource</var>'s origin, return `true`.
-4.  Return `false`.
+5.  Return `false`.
 
-Step 2 returns `true` if the fetch was a CORS-enabled request. If the
+Step 3 returns `true` if the fetch was a CORS-enabled request. If the
 fetch failed the CORS checks, it won't be available to us for integrity
 checking because it won't have loaded successfully.
 {:.note}
 
+Since the [response type][] for data URLs will always be "opaque" for
+`script` and `link` elements, such URLs are never eligible for integrity
+checks. Blob URLs on the other hand are usually considered same-origin and
+therefore are eligible for integrity checks.
+{:.note}
+
 [fetch-mode]: https://fetch.spec.whatwg.org/#concept-request-mode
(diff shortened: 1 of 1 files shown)


mozfreddyb:
The [File API spec](https://w3c.github.io/FileAPI/#originOfBlobURL) says that blobs are bound to the origin that created them, but they are not limited to be *used* within other origins, i.e. one can send blob objects to other origins using postMessage.

To be honest, I do not fully understand whether this means one should be allowed to load cross-origin blobs in scripts (without integrity).

Maybe @annevk can help clarifying this.

annevk:
A blob itself doesn't have an origin and is not bound. But a blob URL is bound to the origin of the global object of the method that was used to create the blob URL.

fmarier:
The note in this PR is what I understood from my email discussion with @annevk and I'm hoping he can review this for us to ensure it's a good summary of what the various specs say about data and blob.

specs/subresourceintegrity/spec.markdown
```diff
@@ -350,19 +350,29 @@ The following algorithm details these restrictions:
 
 1.  Let <var>request</var> be the request that fetched
     <var>resource</var>.
-2.  If the [mode][fetch-mode] of <var>request</var> is `CORS`,
+2.  If the [response tainting][] of <var>request</var> is `opaque`,
+    return`false`.
```
annevk: If resource is the response, why not simply check whether it is opaque? It seems weird to use properties of the request after fetching finished.

annevk:
It seems the eligibility algorithm should just check response's type and that it's "basic", "cors", or "default". Everything else is not needed.

fmarier:
> It seems the eligibility algorithm should just check response's type and that it's "basic", "cors", or "default". Everything else is not needed.

While the CORS and same-origin steps aren't really needed, the current algorithm was quite useful to me while figuring out how to implement this in Firefox, given that we don't use `fetch()` internally for everything yet. So perhaps we should keep it that way in v1 and revisit in SRI v2 when `fetch()` is more tightly integrated in user agents?

I've addressed your [other comment](https://github.com/w3c/webappsec/pull/492#discussion_r41004685) in my last push.

devd:
this looks great to me. we can revert or make more changes if people think this is still a bit unclear.

annevk:
It's not about `fetch()`, it's about what `fetch()` invokes, the "fetch" algorithm. In any event, given that you already define things in terms of the "fetch" algorithm it seems weird to deviate here based on your personal experience with Firefox. E.g., those implementing fetch in Servo will end up confused.

devd:
hmm ..the current algorithm just checks "opaque", "cors", or same-origin. Are you saying we should remove the opaque section?

annevk:
You should check that response's type is "basic", "cors", or "default". That's all.

fmarier:
Let's move the discussion to https://github.com/w3c/webappsec-subresource-integrity/pull/4

PR #496, opened 2015-10-03
Update URLs in README file

deian:


mikewest:
We could probably pull in the table from http://w3c.github.io/webappsec/specs/, WDYT?

deian:
:+1: I like that, will fix

mikewest:
Thanks!

PR #510, opened 2016-02-29
Mark plugin-types as a non-default source

oreoshake:
I'm not sure if this is correct, but I think plugin-types does not inherit from `default-src`. 

/cc @mikewest


mikewest:
1. CSP has moved to https://github.com/w3c/webappsec-csp.
2. `plugin-types` is (currently) split out into https://w3c.github.io/webappsec-csp/document/.
3. I think the text around https://w3c.github.io/webappsec-csp/#directive-default-src pretty clearly scopes the directive to the set of "fetch directives". I'm not sure explicitly listing all the other directives that it doesn't apply to is worthwhile.

WDYT?

oreoshake:
Boy, it's been while since I've looked at the spec but I agree, the `default-src` being limited to "fetch" directives explanation is :ok_hand:. I'm surprised I didn't deduce that from the set of directives that I came up with as being non-default.

PR #559, opened 2019-12-09
Make redirect for SRI spec include fragment ID

sideshowbarker:
This change makes the specs/subresourceintegrity/index.html file’s in-document redirect include any fragment ID in the target URL the request gets redirected to. That prevents any existing links like https://w3c.github.io/webappsec/specs/subresourceintegrity/#HTMLLinkElement (in, eg., MDN articles) from otherwise breaking.

### DIFF
--- a/specs/subresourceintegrity/index.html
+++ b/specs/subresourceintegrity/index.html
@@ -1,7 +1,6 @@
 <!DOCTYPE html>
 <html>
 <head>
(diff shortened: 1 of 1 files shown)


VChernomYA:
specs/subresourceintegrity/index.html

marcoscaceres (approved):

PR #579, opened 2021-04-22
Widen scope of Permission API

marcoscaceres:
We are looking at expanding the scope of Permissions API to have opinions about UI, lifetimes, etc. so the text is overly restrictive.

### DIFF
--- a/admin/webappsec-charter-2021.html
+++ b/admin/webappsec-charter-2021.html
@@ -510,11 +510,7 @@
 		applications to be aware of the status of a given
 		permission, to know whether it is granted, denied, or
 		if the user will be asked whether the permission
-		should be granted. This specification will not
-		address user agent implementations of permissions,
-		including their scope, duration, granularity, or user
-		interface and experience for indicating, configuring,
-		or asking for permissions.</p>
+		should be granted.</p>
 
               <p class="draft-status"><b>Draft state:</b> <a href="https://www.w3.org/TR/2020/WD-permissions-20200720/">Working Draft</a></p>
 


mikewest:
Exciting. Are there sketches of these plans anywhere to skim through? I'm certainly not opposed to doing more work in this area; it's important.

/cc @clelland, @engedy, and @dveditz

marcoscaceres:
Good starting point https://github.com/w3c/geolocation-api/issues/69#issuecomment-805388266

And issue: 
https://github.com/w3c/permissions/issues/233

If we agree to widen the scope I'll start drafting up some ideas.

clelland:
I would love to see the permissions policy features covered by this spec in some way, which could influence the text around UI and lifetimes, but I'm excited to see what comes out of this.

marcoscaceres:
Ok, so, my rough idea is start with adding a section that broadly describes lifetimes or whatever we like. 

Then, in the [Permissions Registry](https://w3c.github.io/permissions/#permission-registry), we add something like: 

**Optional** _UI requirements_:
  - UI requirements for the given permission.

So, if there are requirements, they can be individually added for each permission as needed (same as other things in the spec). 

Then, for example, in the [Geolocation Permission](https://w3c.github.io/permissions/#geolocation) we would add something like:

> The UI requirements for the Geolocation Permission are as follows: ....  [=lifetime=] MUST something something....

VChernomYA (approved):

VChernomYA:
admin/webappsec-charter-2021.html

PR #581, opened 2021-05-06
[charter 2021] allow going to rec; require 2x expected support

samuelweiler:
per discussions on 20 April 2021 WG call: https://github.com/w3c/webappsec/blob/main/meetings/2021/2021-04-20-minutes.md

### DIFF
--- a/admin/webappsec-charter-2021.html
+++ b/admin/webappsec-charter-2021.html
@@ -556,7 +556,7 @@
 	  <h2>Success Criteria</h2>
 	  <p>In order to advance to <a href="https://www.w3.org/Consortium/Process/#RecsPR" title="Proposed Recommendation">Proposed Recommendation</a>, each normative specification is expected to have <a href="https://www.w3.org/Consortium/Process/#implementation-experience">at least two independent implementations</a> of every feature defined in the specification.</p>
 	  <p>Each specification should contain separate sections detailing all known security and privacy implications for implementers, Web authors, and end users.</p>
-  	  <p>All new features should be supported by at least two intents to implement before being incorporated in the specification.</p>
+  	  <p>New features should be expected to be supported by at least two implementations, which may be judged by factors including existing implementations, expressions of interest, and lack of opposition.</p>
 	  <p>There should be testing plans for each specification, starting from the earliest drafts.</p>
 	  <p>
 		Each specification should contain a section on
--- a/admin/webappsec-charter-2021.html
+++ b/admin/webappsec-charter-2021.html
@@ -255,15 +255,16 @@
           Deliverables
         </h2>
 
-        <p>More detailed milestones and updated publication schedules are available on the <a href="https://www.w3.org/groups/wg/webappsec/publications">group publication status page</a>.</p>
-
-	<p><i>Draft state</i> indicates the state of the deliverable
+	<p><i>Draft state</i>, below, indicates the state of the deliverable
 	at the time of the charter approval.</p>
 
+	<p>Current document status is available on the <a href="https://www.w3.org/groups/wg/webappsec/publications">group publication status page</a>.</p>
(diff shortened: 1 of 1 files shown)


admin/webappsec-charter-2021.html
```diff
@@ -255,15 +255,16 @@ <h2>
...
+	<p><i>Draft state</i>, below, indicates the state of the deliverable
 	at the time of the charter approval.</p>
 
+	<p>Current document status is available on the <a href="https://www.w3.org/groups/wg/webappsec/publications">group publication status page</a>.</p>
+	      
 	<p>With the exception of the Mixed Content specification, the
 	  Working Group intends to publish the latest state of their
-	  work as Candidate Recommendation (with Snapshots) and does
-	  not intend to advance their documents to Recommendation.</p>
+	  work as Candidate Recommendation (with Snapshots). The group may advance
```
wseltzer, line 265: Suggest s/may/will/

wseltzer (approved):

admin/webappsec-charter-2021.html
```diff
@@ -555,6 +556,7 @@ <h3>
 	  <h2>Success Criteria</h2>
 	  <p>In order to advance to <a href="https://www.w3.org/Consortium/Process/#RecsPR" title="Proposed Recommendation">Proposed Recommendation</a>, each normative specification is expected to have <a href="https://www.w3.org/Consortium/Process/#implementation-experience">at least two independent implementations</a> of every feature defined in the specification.</p>
 	  <p>Each specification should contain separate sections detailing all known security and privacy implications for implementers, Web authors, and end users.</p>
+  	  <p>All new features should be supported by at least two intents to implement before being incorporated in the specification.</p>
```
mikewest, line 559: So far as I can tell, only Mozilla and Google have a common practice of sending something like "intent to implement/prototype" emails. WHATWG's [policy document says](https://whatwg.org/workstream-policy#living-standard) "Living Standards are intended to include only those features that are likely to be implemented in major browser engines and are suitable for adoption and implementation by other browser engine developers and integrators.", HTML's PR template includes "At least two implementers are interested (and none opposed):", and their [working mode document has some nuanced suggestions](https://whatwg.org/working-mode#changes).

With that in mind, it might be reasonable to frame this a little less narrowly?

admin/webappsec-charter-2021.html
samuelweiler: No objection.

admin/webappsec-charter-2021.html
hober: Came here to say what @mikewest already said.

wseltzer: Would you say instead: "New features should be expected to have at least two implementations, as judged by expressions of interest (and lack of opposition)."?

Another question we've heard is whether features should be dropped after some time window if expected implementations don't materialize. Thoughts?

annevk: Wendy: yes and yes. 😊 (Though that window should be somewhat large, on the order of years.)

dveditz: I agree with the comments from Mike and Anne. The proposed text in the PR matches the spirit of what we intend in shorthand, but might be too blunt and invite rule-lawyering. I like the nuance of the WHATWG working-mode document and would like to capture more of it. 

I imagine it's not allowed to have external definitions in a charter, since they technically could change out from under us. And it's kind of a lot of text to incorporate wholesale (and might need tweaks besides). As a short form the sentence Wendy proposed two comments above is better.

dveditz (changes requested):

dveditz (approved):
Looks good

mikewest (approved):
LGTM2. I'll merge it in.

PR #583, opened 2021-05-14
Additional clarification

wseltzer:


wseltzer:
a colleague was confused by the reference to Mixed Content, so attempting to clarify.

PR #588, opened 2021-06-01
Clarify that new feature limit applies only once a spec hits CR

samuelweiler:
Per suggestion from @dontcallmedom

### DIFF
--- a/admin/webappsec-charter-2021.html
+++ b/admin/webappsec-charter-2021.html
@@ -550,10 +550,11 @@
       </section>
 
 	<section id="success-criteria">
-	  <h2>Success Criteria</h2>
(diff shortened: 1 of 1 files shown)


wseltzer (approved):

dveditz (approved):

annevk:
Doesn't this mean that everything that's in the initial CR might lack this? This seems like an incentive to not go to CR or instead of maintaining a CR, start a new document.

sideshowbarker (approved):

mikewest:
My view is that shifting from WD to CR would imply the kind of commitments that we'd ask for when changing a CR. Assuming Mozilla, Apple, Google, Microsoft, Brave, etc. remain members of the WG, that seems possible. Perhaps we could make that explicit?

samuelweiler:
How's this? 

"In order to advance to Candidate Recommendation and to add features after reaching Candidate Recommendation, each feature is expected to be supported by at least two implementations, which may be judged by factors including existing implementations, expressions of interest, and lack of opposition." 

And, as before:

> In order to advance to Proposed Recommendation, each normative specification is expected to have at least two independent implementations of every feature defined in the specification.

PR #592, opened 2021-06-08
new charter start date TBD

xueyuanjia:


samuelweiler:
I will make the same change in the [static snapshot for the AC review](https://www.w3.org/2021/06/webappsec-proposed-charter-2021.html).

xueyuanjia:
> I think you meant to change the date to "TBD"..

Yes. Thanks @samuelweiler

PR #594, opened 2021-06-10
A couple more COI-adjacent items

arturjanc:


mikewest:
Perhaps we can explicitly link to @shhnjk's analysis of the profiling API?

arturjanc:
Hah, I wrote this before seeing Jun's reply. Good point, added!

mikewest:
Thanks!

PR #609, opened 2022-04-22
Add 'A Well-Known URL for Changing Passwords' to the WG homepage.

hober:


VojtaStanek:
Hi @mikewest!

Is this PR enough to get in on https://www.w3.org/2011/webappsec/? I haven't found this addition there yet. There are two `index.html` files in this repository - one in the root of the repository and another in the `homepage` directory and I am not sure which one is the correct one. 

Thanks for this and your work making the web more secure!

mikewest:
Hrm. No, it isn't.

@samuelweiler and @wseltzer might know where that page is controlled; I'll happily update it if I have access.

samuelweiler:
More importantly, "A Well-Known URL for Changing Passwords" doesn't appear on https://www.w3.org/groups/wg/webappsec/publications, and the way to fix that is not by editing the page but instead by issuing a First Public Working Draft.   @hober, are you up for that?

As to the WebAppSec homepage, we're going to deprecate the bespoke page at https://www.w3.org/2011/webappsec/ in favor of the more-automated https://www.w3.org/groups/wg/webappsec/.  You should see that change shortly.

(The bespoke page is still hosted in CVS.  We once had ambitions of moving it to Github, hence this copy of the data in GitHub, but the authoritative source is still - just for a few more moments - CVS.)

hober:
IIRC the group agreed to publish a FPWD ages ago.

PR #650, opened 2024-04-04
Revert "Drop e2e (#649)"

### LINKED ISSUE #646: "End-to-End Encryption email" is missing an actual proposal
From [AC Review](https://lists.w3.org/Archives/Public/public-review-comments/2024Mar/0000.html):
[[
We are concerned about the new "End-to-End Encryption email" proposed optional deliverable because it looks out of scope for the working group: https://www.w3.org/2024/01/proposed-wg-webappsec.html#scope
* The proposed charter does not have cryptography in its scope except for maintaining the Web Cryptography API
* The proposed charter does not have messaging standards, formats, protocols in its scope

We also note that the link in the deliverables section to "End-to-End Encryption email" is also only to a section of minutes of a meeting from nearly 6 months ago, and NOT a Proposal draft (whether incubated or not) or even an Explainer draft, which is highly unusual and unexpected. This seems like a Charter drafting clerical error of leaving in the wrong link, since the linked minutes do say there is a "draft that we have" that "if there's enough interest we will publish", so presumably that publication may (should?) have happened in the past 6 months since one implementer proposed it and another implementer said "There is interest".

Since we are not sure if there was an attempt to expand the Scope but it wasn't written up, and/or if there was a clerical error of failing to link to an actual proposal, or if this was perhaps a Charter editing-in-progress tentative addition that was errantly not removed before taking to an AC poll, we do not feel we have enough information to propose a specific set of actions to resolve these concerns.

We are expressing our concerns (on apparent scope violation and failure to link to an actual proposal) but we are not making this a Formal Objection because optimistically the error or errors may again be clerical (rather than substantial) in nature, and we expect the Team to address such corrections before adopting an updated Web Application Security Working Group Charter.
]]

cc https://github.com/w3c/strategy/issues/426

marcoscaceres:
This reverts commit 14fe7372f6b0a8ee78b7292b66ae52884d5b1c25.

Proposals is here:
https://github.com/WebKit/explainers/blob/main/cryptographic-message-syntax-API/README.md

And works in conjunction with:
https://github.com/WebKit/explainers/tree/main/remote-cryptokeys

Closes #646

### DIFF
--- a/admin/webappsec-charter-2023.html
+++ b/admin/webappsec-charter-2023.html
@@ -556,6 +556,11 @@
           following documents:
         </p>
         <dl>
+          <dt class="spec"><a href="https://github.com/w3c/webappsec/blob/main/meetings/2023/2023-09-14-TPAC-minutes.md#end-to-end-encryption">End-to-End Encryption email</a></dt>
+          <dd>
+              This idea would allow end-to-end encryption of emails, without exposing key content,
+              in Web applications.
+          </dd>
           <dt class="spec"><a href="https://github.com/w3c/webappsec/blob/main/meetings/2023/2023-09-15-TPAC-minutes.md#source-code-transparency-sketch">Source Code Transparency</a></dt>
           <dd>
             The goal would be to have a mechanism to verify that the source code of a web app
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Ok, it would be great to add it back in the near future.

